PR #7, opened 2021-03-03
Strip refs/tags prefix from GITHUB_REF in workflows

weierophinney:
Evidently `$GITHUB_REF` is in the form `refs/tags/{TAG_NAME}` for a release.

### DIFF
--- a/.github/workflows/build-and-push-containers.yml
+++ b/.github/workflows/build-and-push-containers.yml
@@ -13,10 +13,11 @@ jobs:
       - name: Compile tag list
         id: tags
         run: |
+          TAG=${GITHUB_REF/refs\/tags\//}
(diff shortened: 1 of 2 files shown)


.github/workflows/build-and-push-containers.yml
```diff
@@ -13,10 +13,11 @@ jobs:
       - name: Compile tag list
         id: tags
         run: |
+          TAG=${GITHUB_REF/refs\/tags\//}
```
glensc, line 16: You should use strip prefix in such cases;

```
➔ echo TAG=${GITHUB_REF/refs\/tags\//}
TAG=123
➔ echo TAG=${GITHUB_REF#refs/tags/}
TAG=123
```

PR #12, opened 2021-03-04
Provide ability to provide checks to run in addition to auto-discovered checks

### LINKED ISSUE #11: Add the ability to specify additional checks to perform
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | yes
| BC Break    | no

#### Summary

Currently, the `.laminas-ci.json` file allows you to specify a "checks" key, which is an array of checks to run. When defined, it specifies your _entire_ matrix to run; the action does no discovery of its own at that point.

Sometimes, however, you may only want to **add** one or more checks to perform _in addition to_ any that were discovered. As an example, this could solve the problem presented in #4, which requests the ability to run code coverage reports, or a scenario where you want to run an integration test suite that's normally excluded from your unit tests.

My proposal is to allow the following additions to the `.laminas-ci.json` format:

```json
{
    "additional_checks": [
        {}
    ]
}
```

where each object in the `additional_checks` array has the following format (same format as individual items in the matrix):

```json
{
      "name": "(string) Name of the check being run",
      "operatingSystem": "(string) Name of the OS the job should be run on (generally ubuntu-latest)",
      "action": "(string) GHA to run the step on; currently ignored, as GHA does not support dynamic action selection",
      "job": "(string) JSON object detailing the job"
}
```

Only "name" and "job" would be required; the other items could use defaults.

The "job" would then follow the same format as jobs currently do:

```json
{
  "php": "string PHP minor version to run against",
  "extensions": [
    "extension names to install; names are from the ondrej PHP repository, minus the php{VERSION}- prefix",
  ],
  "ini": [
    "php.ini directives, one per element; e.g. 'memory_limit=-1'",
  ],
  "dependencies": "dependencies to test against; one of lowest, locked, latest",
  "command": "command to run to perform the check",
}
```

with the following rules:

- "php" would be optional; if not provided, the assumption would be to use the "stable-php" version
  - Nice to have: "*" to mean run against all versions; the check gets cloned, with each version using a different PHP version
-  "extensions" is optional; if not provided, an empty array will be used
- "ini" is optional; if not provided, an empty array will be used
- "dependencies" is optional; if not provided, the assumption is to use "locked"
  - Nice to have: "*" to mean run against each of latest, locked, and lowest
  - Nice to have: allow an array of values; if an array is present, one job is created per PHP version
- "command" is required

This would give consumers more flexibility in defining one-off checks, or to run checks that we do not yet have the ability to detect in the container.

weierophinney:
Previously, usage of the "checks" array would cause usage of that value for the matrix, and no further discovery would occur.  This patch adds a new key, `additional_checks`, which allows package authors to provide checks to run _in addition to any discovered_.  This allows providing checks for tools the matrix discovery tools do not know about, or providing one-off checks (such as benchmarks or mutation tests).

The syntax for the `additional_checks` key is as follows:

```json
{
    "additional_checks": [
        {
            "name": "(string; REQUIRED) name of the check to run",
            "job": {
                "command": "(string; REQUIRED) command to run",
                "php": "(string; OPTIONAL) PHP version to run on (defaults to stable-php); * indicates all versions",
                "dependencies": "(string; OPTIONAL) dependency set to run on (defaults to locked); * indicates all sets",
                "extensions": [
                    "(array of strings; OPTIONAL) specific extensions to install for this check only"
                ],
                "ini": [
                    "(array of strings; OPTIONAL) specific php.ini settings to use for this check only"
                ]
            }
        }
    ]
}
```

A job per PHP version per dependency set will be created, and the "name" will be appended with "on PHP {VERSION} with {DEPS} dependencies" during an actual run.

The tool discovers checks first, then appends any `additional_checks` are concatenated, and then any `exclude` rules are applied.

Fixes #11

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,6 +108,10 @@ The package can include a configuration file in its root, `.laminas-ci.json`, wh
     {
     },
   ],
+  "additional_checks": [
+    {
+    },
+  ]
   "exclude": [
     {
     }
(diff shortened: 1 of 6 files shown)


gsteel:
Awesome! I'm gonna go try that out as soon as I can. Thanks @weierophinney

PR #25, opened 2021-04-08
No checks job must have the configured extensions and ini

Slamdunk:
See https://github.com/laminas/laminas-cache-storage-adapter-apcu/pull/6/checks?check_run_id=2295992531

Even on `No checks` job the dependencies are installed, so the extensions and ini settings configured by `.laminas-ci.json` file must be in place

### DIFF
--- a/src/create-jobs.js
+++ b/src/create-jobs.js
@@ -67,8 +67,8 @@ const createNoOpJob = function (config) {
         JSON.stringify(new Command(
             "echo 'No checks discovered'",
             config.stable_version,
-            [],
-            [],
+            config.extensions,
+            config.php_ini,
             'locked',
         )),
     )];


boesing:
Yup, I've seen the same. I'd prefer going the other way and avoid to checkout or install composer dependencies at all.
For this, I've created two PRs:

https://github.com/laminas/laminas-continuous-integration-action/pull/23
https://github.com/laminas/laminas-ci-matrix-action/pull/26

But #26 introduces a BC break. Lets wait for Matthews review to see if the BC break is a real problem as the GHA always checks out the latest versions of the actions.

boesing:
I've applied a non-breaking change with an empty command and thus added two new PRs:

https://github.com/laminas/laminas-continuous-integration-action/pull/24
https://github.com/laminas/laminas-ci-matrix-action/pull/27

But I think we can apply this PR anyways as its not wrong either.

weierophinney:
Closing in favor of #27 and laminas/laminas-continuous-integration-action#24, as that combination does not involve a BC break, and does not require aggregating INI and extension information (as it leads to a no-op).

PR #26, opened 2021-04-08
feature: use null for no check

boesing:
|    Q          |   A
|-------------- | ------
| BC Break      | yes
| New Feature   | yes

### Description

Pass `null` in case there is no need to execute a command. This is more accurate due to the fact that there is no need to checkout or install any source code or `composer` deps.

In https://github.com/laminas/laminas-continuous-integration-action/pull/23 I've implemented the handling of the `null` command.

Due to the fact that this may return `null`, this supposed to be a BC break. Should we create a new major for this?

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,10 +61,10 @@ It spits out a single output, "matrix", which is a JSON string in the following 
       "name": "(string) Name of the check being run",
       "operatingSystem": "(string) Name of the OS the job should be run on (generally ubuntu-latest)",
       "action": "(string) GHA to run the step on; currently ignored, as GHA does not support dynamic action selection",
-      "job": "(string) JSON object detailing the job (more on this later)",
-    },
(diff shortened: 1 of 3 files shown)


weierophinney:
This is tricky in terms of figuring out BC.

On the one hand, we're expanding what is returned for a "command" value. Return expansion is generally considered a BC break in terms of APIs, as it means the consuming code might be operating on a value it does not understand.

On the flip side... we're likely the only consumers, so as long as we have the laminas-continuous-integration-action updated, it shouldn't matter.

However, I have another proposition here: why not, instead of returning null here, or returning the original "echo" command, we return an empty string? That keeps the value type (and thus no BC break), and then we can update the laminas-continuous-integration-action to check for an _empty_ command, and end execution early in that scenario.

boesing:
I'll close this in favor of a new PR to use an empty command instead. Thanks for pointing this out!

PR #27, opened 2021-04-08
feat: use empty command for no check

boesing:
|    Q          |   A
|-------------- | ------
| BC Break      | no
| New Feature   | yes

### Description

Pass the command as an empty string in case there is no need to execute a command. This is more accurate due to the fact that there is no need to checkout or install any source code or `composer` deps.

In https://github.com/laminas/laminas-continuous-integration-action/pull/24 I've implemented the handling of the empty string command.

With the current version of `laminas-continuous-integration-action`, the check will fail as the action will exit with status `1`.


weierophinney (approved):
:ship:

Issue #24, opened 2021-04-02
Pull requests without dedicated branch fail creating the matrix


boesing:
### Bug Report

|    Q        |   A
|------------ | ------
| Version(s)  | 1.5.2

#### Summary

It seems that for pull requests, which are created from the same branch in the forked repository, it is not possible to create a diff for the matrix generation.

#### Current behavior

Target branch: `1.2.x`
Project branch: `1.2.x`

```
Cloning into '.'...
Cloning repository
Checking out branch 1.2.x
Already on '1.2.x'
Your branch is up to date with 'origin/1.2.x'.
Fetching target ref refs/pull/7/merge
fatal: Refusing to fetch into current branch refs/heads/1.2.x of non-bare repository
```

https://github.com/laminas/laminas-mvc-i18n/pull/7/checks?check_run_id=2254661339

#### How to reproduce

Fork a repository, do not create a dedicated feature/bugfix/whatever branch but push to the same branch, the PR is targeted to later on.

#### Expected behavior

Matrix generation works without conflicting when creating a diff.
Works when the source branch differs: https://github.com/laminas/laminas-mvc-i18n/pull/10/checks?check_run_id=2254711601

boesing:
Uh, interesting fact. I've created a dedicated PR with the same commit but other source branch and the same checks are applied for both PRs:

https://github.com/laminas/laminas-mvc-i18n/pull/7
https://github.com/laminas/laminas-mvc-i18n/pull/10

gpgkd906:
@boesing 
I think I have meet the same problem, with this [PR](https://github.com/laminas/laminas-permissions-rbac/pull/13)
how can i fix this?

boesing:
@gpgkd906 Best practice is to have a dedicated branch for anything you are working on in git.
So you should create a feature branch, e.g. `feature/php80`, and use that for PRs.

You can use this cheat sheet for more details on how to create that branch: https://education.github.com/git-cheat-sheet-education.pdf

gpgkd906:
@boesing 
Thanks very much.
I create a feature branch ```feature/php80 ``` and create a new [PR](https://github.com/laminas/laminas-permissions-rbac/pull/20) then it passed.
the interesting things is, when new PR passed, the old [PR](https://github.com/laminas/laminas-permissions-rbac/pull/13) which was failed, now is passed too.

boesing:
[...]

PR #29, opened 2021-04-20
Ensure pull request branch name does not cause collisions

weierophinney:
When a pull request is made by a user and they name their branch the same name as the target branch against which the pull request is being made, we run into a conflict, which causes the checkout to fail (see https://github.com/mezzio/mezzio-helpers/pull/7/checks?check_run_id=2393994090#step:3:9).

To resolve this scenario, if we detect a pull_request, we can give the local branch name a prefix, `pull/`.

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -5,11 +5,14 @@ set -e
 function checkout {
     local REF=
     local LOCAL_BRANCH=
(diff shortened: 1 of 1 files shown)


boesing:
Fixes #24

PR #31, opened 2021-05-10
Use `XmlLint` against tool configurations (PHPUnit, PHPCS, Psalm)

internalsystemerror:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | yes

### Description

This PR adds the use of `xmllint` if added via https://github.com/laminas/laminas-continuous-integration-action/pull/37

Note: I don't personally use psalm, and have not checked that the xmllint command works for this schema. However I obtained the schema from the `psalm.xml` of `mezzio/mezzio`, as I did with PHPUnit and PHPCS, both of which work correctly.

### DIFF
--- a/src/action/github.ts
+++ b/src/action/github.ts
@@ -1,9 +1,7 @@
+import * as core from '@actions/core';
 import {Action} from '../action';
 import {Output} from '../config/output';
 import {Logger} from '../logging';
-
-/* eslint-disable-next-line import/no-commonjs, @typescript-eslint/no-var-requires */
-const core = require('@actions/core');
 
 export class Github implements Action {
     publish(variable: string, output: Output): void {
--- a/package.json
+++ b/package.json
@@ -20,6 +20,7 @@
   },
   "scripts": {
     "lint": "eslint src/ --ext .ts",
+    "lint-fix": "eslint src/ --ext .ts --fix",
     "build": "tsc --build && webpack --mode=production"
   },
   "dependencies": {
--- a/src/config/app.ts
+++ b/src/config/app.ts
@@ -54,6 +54,7 @@ export interface JobDefinition {
     composerDependencySet: ComposerDependencySet;
     ignorePhpPlatformRequirement: boolean;
     additionalComposerArguments: string[];
+    beforeScript: string[];
 }
 
 export interface Job {
@@ -150,7 +151,8 @@ function convertJobDefinitionFromFileToJobDefinition(
         job.extensions ?? config.phpExtensions,
         job.ini ?? config.phpIni,
         discoverIgnorePhpPlatformRequirementForJobByVersion(job, phpVersion, config),
-        discoverAdditionalComposerArgumentsForCheck(job, config)
+        discoverAdditionalComposerArgumentsForCheck(job, config),
+        job.before_script
     );
 }
 
@@ -161,7 +163,8 @@ function createJobDefinition(
     phpExtensions: string[],
     phpIniSettings: string[],
     ignorePlatformRequirements: boolean,
-    additionalComposerArguments: string[]
+    additionalComposerArguments: string[],
+    beforeScript: string[],
 ): JobDefinition {
     return {
         php                          : phpVersion,
@@ -170,7 +173,8 @@ function createJobDefinition(
         phpIni                       : phpIniSettings,
         composerDependencySet        : composerDependencySet,
         ignorePhpPlatformRequirement : ignorePlatformRequirements,
-        additionalComposerArguments  : additionalComposerArguments
+        additionalComposerArguments  : additionalComposerArguments,
+        beforeScript                 : beforeScript,
     };
 }
 
@@ -283,12 +287,14 @@ function createJobsForTool(
     tool: Tool
 ): JobFromTool[] {
     const jobs: JobFromTool[] = [];
+    const beforeScript: string[] = tool.lintConfigCommand
+        ? tool.filesToCheck.map((file) => `${tool.lintConfigCommand} ${file}`)
+        : [];
 
     if (tool.executionType === ToolExecutionType.STATIC) {
         const lockedOrLatestDependencySet: ComposerDependencySet = config.lockedDependenciesExists
             ? ComposerDependencySet.LOCKED
             : ComposerDependencySet.LATEST;
-
 
         return [
             createJob(
@@ -300,7 +306,8 @@ function createJobsForTool(
                     config.phpExtensions,
                     config.phpIni,
                     config.ignorePhpPlatformRequirements[config.minimumPhpVersion] ?? false,
-                    config.additionalComposerArguments
+                    config.additionalComposerArguments,
+                    beforeScript,
                 ),
                 tool
             ) as JobFromTool
@@ -318,7 +325,8 @@ function createJobsForTool(
                     config.phpExtensions,
                     config.phpIni,
                     config.ignorePhpPlatformRequirements[config.minimumPhpVersion] ?? false,
-                    config.additionalComposerArguments
+                    config.additionalComposerArguments,
+                    beforeScript,
                 ),
                 tool
             ) as JobFromTool);
@@ -332,7 +340,8 @@ function createJobsForTool(
                 config.phpExtensions,
                 config.phpIni,
                 config.ignorePhpPlatformRequirements[version] ?? false,
-                config.additionalComposerArguments
+                config.additionalComposerArguments,
+                beforeScript,
             ), tool) as JobFromTool,
 
             createJob(tool.name, createJobDefinition(
@@ -342,7 +351,8 @@ function createJobsForTool(
                 config.phpExtensions,
                 config.phpIni,
                 config.ignorePhpPlatformRequirements[version] ?? false,
-                config.additionalComposerArguments
+                config.additionalComposerArguments,
+                beforeScript,
             ), tool) as JobFromTool,
         ));
     }
@@ -368,6 +378,7 @@ function createNoOpCheck(config: Config): Job {
             phpIni                       : [],
             ignorePhpPlatformRequirement : config.ignorePhpPlatformRequirements[config.stablePhpVersion] ?? false,
             additionalComposerArguments  : config.additionalComposerArguments,
+            beforeScript                 : [],
         }
     };
 }
--- a/src/config/input.ts
+++ b/src/config/input.ts
@@ -11,6 +11,7 @@ export interface ConfigurationFromFile {
     ignore_php_platform_requirements?: IgnorePhpPlatformRequirements;
     stablePHP?: string;
     additional_composer_arguments?: string[];
+    before_script: string[];
 }
 
 export interface JobExclusionsFromFile {
@@ -77,6 +78,7 @@ export interface JobDefinitionFromFile {
     ignore_php_platform_requirement?: boolean;
     additional_composer_arguments: string[];
     command: string;
+    before_script: string[];
 }
 
 export type AnyComposerDependencySet = typeof WILDCARD_ALIAS;
--- a/src/config/input.ts
+++ b/src/config/input.ts
@@ -11,7 +11,6 @@ export interface ConfigurationFromFile {
     ignore_php_platform_requirements?: IgnorePhpPlatformRequirements;
     stablePHP?: string;
     additional_composer_arguments?: string[];
-    before_script: string[];
 }
 
 export interface JobExclusionsFromFile {
--- a/src/config/output.ts
+++ b/src/config/output.ts
@@ -11,6 +11,7 @@ export interface JobDefinitionForMatrix
     ignore_platform_reqs_8: boolean, // eslint-disable-line camelcase
     ignore_php_platform_requirement: boolean, // eslint-disable-line camelcase
     additional_composer_arguments: Array<string> // eslint-disable-line camelcase
+    before_script: string[]; // eslint-disable-line camelcase
 }
 
 export interface JobForMatrix {
@@ -47,7 +48,9 @@ export function createJobForMatrixFromJob(job: Job): JobForMatrix {
             /* eslint-disable-next-line camelcase */
             ignore_php_platform_requirement : job.job.ignorePhpPlatformRequirement,
             /* eslint-disable-next-line camelcase */
-            additional_composer_arguments   : job.job.additionalComposerArguments
+            additional_composer_arguments   : job.job.additionalComposerArguments,
+            /* eslint-disable-next-line camelcase */
+            before_script                   : job.job.beforeScript,
         }
     };
 }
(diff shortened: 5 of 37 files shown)


src/create-jobs.js
```diff
@@ -82,7 +82,34 @@ const createNoOpJob = function (config) {
  * @return {Array}
  */
 function checks (config) {
+    const xmlSchema = {
+        phpunit: 'https://schema.phpunit.de/9.3/phpunit.xsd',
```
froschdesign: This may fail if a schema version other than 9.3 is used, e.g. 9.5.

internalsystemerror: I considered that, ideally it would just the schema defined in the XML. Feel free to take over this PR if you have some ideas. I considered relying on the xsd installed into vendor. I'm fairly certain PHPUnit would work that way, but I'm not sure about Psalm. Does that come with an XSD installed to vendor?

internalsystemerror: Pushed a change to use the schemas installed to vendor

weierophinney (changes requested):
My primary concern with this is that if these jobs fail... we've still queued jobs for the tools referenced, which means that they still run, but with known invalid schemas.

I think it might make more sense to inline the xmllint into the commands: `xmllint ... && command ...`. This means that we fail early if the config is incorrect for the jobs running those tools.

internalsystemerror:
Yeah I was looking into being able to set a job as dependent on another as I considered that a better approach. Do you have any ideas on that? Otherwise I will just implement the changes you requested.

internalsystemerror:
The problem with inlining the commands is what/how to do that with the PHPUnit matrix? That's why I considered a better solution to be allowing the jobs to be dependent on other jobs.

weierophinney:
@internalsystemerror I think it may make more sense to validate _in this action_; you can see how @boesing managed it for the `composer.json` in #33 . I think there's similar tools for JS?

weierophinney:
> I think it may make more sense to validate _in this action_

On second thought, maybe not - many of these reference a schema file that's installed with the tooling, which happens via Composer.

So, another possibility: do the linting _as a precursor_ to running the associated command?

internalsystemerror:
~~@weierophinney As it stands, the linting will run as a precursor to the associated commands since the jobs will be created with `xmllint ... && command`.~~ EDIT: This was incorrect.

I've fetched the latest changes since this was made, too.

internalsystemerror:
I have modified the commands to run `xmllint ... && command`.

boesing:
Not 100% sure but if a project has both dist and non-dist, this would no create two jobs, right? 

Plenty of unrelated changes there as well. 😅

internalsystemerror:
> Not 100% sure but if a project has both dist and non-dist, this would no create two jobs, right?

Correct, I believe it would.

> Plenty of unrelated changes there as well. 😅

Sorry, I have format on save. I've corrected these but left in the couple of syntax fixes it added.

boesing:
I wonder if it would be possible to avoid having two files created. 
Maybe with using `test` to check if file exists before linting? 

Pseudo-Code:
```
((test -r phpunit.xml && xmllint phpunit.xml ...) || (test -r phpunit.xml.dist && xmllint phpunit.xml.dist)) && vendor/bin/phpunit
```

internalsystemerror:
Nice idea, that would avoid 2 jobs being created if 2 files exist. Thinking about it though, if a second file is submitted (dist xml along with xml), either this is accidental, in which case having the same job twice would be an indication. Or, it's deliberate, in which case you may wish to run these separately? Honestly I don't see this latter one happening though.

boesing:
> Or, it's deliberate, in which case you may wish to run these separately? Honestly I don't see this latter one happening though.

The only thing which came to my mind would be a `.laminas/pre-install.sh` script which applies changes to a config temporarily for whatever case.
In combination with [this action](https://github.com/stefanzweifel/git-auto-commit-action), it could lead to issues when manipulating the `.dist` (as it is committed) so you would probably copy the `*.dist` file to the non-dist version, apply your changes you want to have for CI (for whatever reasons) and then execute the tool. Doing it this way would prevent the github action to detect a change (as usually the non-dist files are added to `.gitignore`).

Fun fact - we probably going to do exactly this for PHPUnit regarding the recent changes for deprecations to exceptions: https://github.com/laminas/laminas-continuous-integration-action/pull/54

So in fact, we are not creating a new file from the `.dist` but modifying whatever file will be used for phpunit (which is non-dist prior dist when it comes to priority).

src/create-jobs.js
```diff
@@ -119,35 +125,92 @@ function checks (config) {
     return [
         new Check(
             config.code_checks,
-            [fileTest('phpunit.xml.dist'), fileTest('phpunit.xml')],
+            [fileTest('phpunit.xml.dist')],
```
boesing: not sure if this is a good one.
This will probably create two checks in projects which do have both files.

----

And again, its not about "how likely that will happen". I've provided an [example](https://github.com/laminas/laminas-ci-matrix-action/pull/31#issuecomment-934876500) on why this could happen and thus, this is not that uncommon.

internalsystemerror: The scope of this was just to run `xmllint` against the various xml files.

I've just been cleaning up old PRs and issues and saw this was still outstanding. I'm unsure what change you want me to make here. If there should be a new script which alters how the .xml vs .xml.dist files are handled, I would suggest that should be a new issue with its own PR, and this blocked until that one is merged and rebased after? If that is sufficient, I can create a new PR to do that when I get some time (won't be soon).

src/create-jobs.js
```diff
@@ -119,35 +125,92 @@ function checks (config) {
...
+            /**
+             * @param {Config} config
+             * @return {Array}
+             */
+            function (config) {
+                return createPHPUnitJobs(
+                    `xmllint --noout --schema ${xmlSchema.phpunit} phpunit.xml.dist`
+                    + './vendor/bin/phpunit',
```
internalsystemerror: These commands need to be concatenated with `&&`

src/create-jobs.js
```diff
@@ -8,19 +8,31 @@ import create_additional_jobs from './additional-checks.js';
...
+        return fs.existsSync(filename);
+
     };
 };
 
+/**
+ * @param {String} filename
+ * @return {String}
+ */
+const appendDistIfNecessary = function(filename) {
```
boesing: can we rename this to `appendDistExtensionIfFileExists` instead?

boesing (changes requested):
LGTM. Just a lil nitpick but from my perspective, this could work. 
@weierophinney do you see anything else here? I think it would be fine if we spawn the Jobs and let them fail then since we need any feedback from the tooling. 
Just not sure about the fact if we should provide the output from the linter to help devs fixing the problem

src/create-jobs.js
internalsystemerror: yep, I'm always bad with names so happy to yield on this.

boesing: same here, but I learned from the best.
/cc @weierophinney 😁

internalsystemerror: My guess is the naming suggestion is to be more descriptive so I went with `appendDistIfFilenameDoesntExist` to clarify that the non-dist version is preferred if both exist.

boesing (approved):

src/create-jobs.js
```diff
@@ -8,19 +8,31 @@ import create_additional_jobs from './additional-checks.js';
...
+const xmlSchema = {
+    phpunit: 'vendor/phpunit/phpunit/phpunit.xsd',
+    phpcs: 'vendor/squizlabs/php_codesniffer/phpcs.xsd',
+    psalm: 'vendor/vimeo/psalm/config.xsd',
+};
+
 /**
  * @param {String} filename
- * @return {Boolean}
+ * @return {function(): Boolean}
```
weierophinney: Is this correct? fs.existsSync is synchronous, so we'd just get back the return value, wouldn't we?

weierophinney: Never mind... just saw we've ALWAYS returned a function that returns boolean. :blush:

src/create-jobs.js
```diff
@@ -78,10 +90,11 @@ const createPHPUnitJobs = function (config) {
  * @return {Job}
  */
 const createPHPUnitJob = function (version, deps, config) {
+    const fileName = appendDistExtensionIfFilenameDoesntExist('phpunit.xml');
```
weierophinney: Change the name of the constant to "configFileName" or "phpunitConfigFile"; "fileName" is to ambiguous.

src/create-jobs.js
```diff
@@ -78,10 +90,11 @@ const createPHPUnitJobs = function (config) {
...
     return new Job(
         'PHPUnit on PHP ' + version + ' with ' + deps + ' dependencies',
         JSON.stringify(new Command(
-            './vendor/bin/phpunit',
+            `xmllint --noout --schema ${xmlSchema.phpunit} ${fileName} && ./vendor/bin/phpunit`,
```
weierophinney: Why `--noout`? It seems like we may want to see _why_ it failed.

src/create-jobs.js
```diff
@@ -136,7 +149,8 @@ function checks (config) {
              * @return {Array}
              */
             function (config) {
-                return createQaJobs('./vendor/bin/phpcs -q --report=checkstyle | cs2pr', config);
+                const filename = appendDistExtensionIfFilenameDoesntExist('phpcs.xml');
```
weierophinney: Same comment as previously: use something like "phpcsConfigFile" or "configFile".

src/create-jobs.js
```diff
@@ -147,7 +161,8 @@ function checks (config) {
              * @return {Array}
              */
             function (config) {
-                return createQaJobs('./vendor/bin/psalm --shepherd --stats --output-format=github --no-cache', config);
+                const filename = appendDistExtensionIfFilenameDoesntExist('psalm.xml');
```
weierophinney: "psalmConfigFile"

weierophinney (changes requested):
Two categories of changes:

- Instead of `const filename = appendDistExtensionIfFilenameDoesntExist(...)`, name the constant something that is semantically appropriate: phpunitConfigFile, phpcsConfigFile, etc.
- I'm thinking we shouldn't use the `--noout` flag. If we have `--noout` and xmllint fails, it's not immediately obvious what happened. With output, the user will know what xmllint errored on.

src/create-jobs.js
internalsystemerror: The docs suggest this only suppresses output, not errors, but I will remove anyway.

src/create-jobs.js
internalsystemerror: Yeah, there are a couple of fixes I made to the doc types in this manner just because I noticed them.

boesing:
The CI container now supports `before_script` and `after_script`.
We could add the linting as one of the `before_script` entries (its a list of commands).

internalsystemerror:
I've been looking at reimplementing this today, using the new `before_script` option. I was mainly looking for a way to only run the xml linting once per file. For example, using `before_script` on PHPUnit, it would run `xmllint` for each `phpunit` job in the matrix.

Does anyone have any ideas on that?

Ocramius:
Wouldn't it be better to run an independent matrix job that only runs xmllint/yamllint for root level xml/yaml files? 🤔

A completely independent failure would be nicer than mixing it with test results, or worse with no tests run at all.

internalsystemerror:
> A completely independent failure would be nicer than mixing it with test results, or worse with no tests run at all.

Preventing tests from running with a known incorrect schema was requested in this comment https://github.com/laminas/laminas-ci-matrix-action/pull/31#pullrequestreview-662187929.

Adding in these checks as just another linting check would certainly be the easiest solution to implement.

weierophinney:
> I was mainly looking for a way to only run the xml linting once per file.

I think the only way to do that is if we were to do this as one or more separate jobs in the workflow, marked to fail the workflow if it fails. Since we can extend workflows now, we could do this in the main workflow in the .github repo...

boesing:
I am not sure if its smart to have linting for configs seperately. I would expect all Jobs related to that Tool failing if the config is invalid rather than having another Job just to lint. 

So I would be +1 for adding these lint checks as a `before_scripts` entry.

Ocramius:
Works for me: no hard opinion, as long as linting manages to make it through, at least :+1:

internalsystemerror:
OK. Then since I have no preference either, I'll update this PR to simply add xmllint as a `before_script` check for each tool, hopefully this weekend.

internalsystemerror:
Forgive the incoming update spam. This took much longer than I want to admit, mainly because I can't get a reliable way to run the tests even on a clean branch.

Pushing now to get test results via GHA, expect fixes (which is updating the expected `matrix.json` to include the new `before_script` commands) shortly.

internalsystemerror:
Checks are green. Ready for review. @boesing tagging you specifically as you had some ideas already for this.

Matrices that are generated for tools will include `before_script: []` (and the XML lint command where appropriate). Those jobs which are set as explicit/additional checks within `.laminas-ci.json` do not include this feature as I was having issues so decided it was out of scope 😄 .

boesing:
I hope I find some time this weekend to review this.
I wonder if we need some generic stuff like `configurationType: XML|YAML|JSON` so that `xmllint` or `yamllint` does not need to get repeated.

And there is still that "if phpunit.xml exists, we should lint that file, else phpunit.xml.dist". But I would assume not to lint both at the same time for example. But I have no clue how to implement that properly, as the `phpunit.xml` might get created within the container. I did that somewhere to actually dynamically set `convertDeprecationsToExceptions` in PHPUnit Configuration to `false`. Maybe @Ocramius can remember that PR as he reviewed that 😂

Ocramius:
https://www.youtube.com/watch?v=DFdDNOnGjWc

internalsystemerror:
It's been waiting 15 months, another few weeks or even months wouldn't hurt it. The approach I took was to KISS. The tools are modified to reduce the `filesToCheck: string[]` to contain only the files that exist, with empty tools being removed.

Functionally it's the same as before, however the ability to add a tool in the future without any files to check has been removed. Then for each file to check that remains, if there is a linting command available it will add it as a `before_script`.

Any more than that I found myself overengineering something that might not need to exist. For example, I'd also like to extend the before_script functionality to include explicit checks that have their own.

src/action/github.ts
```diff
@@ -1,10 +1,8 @@
+import * as core from '@actions/core';
 import {Action} from '../action';
 import {Output} from '../config/output';
 import {Logger} from '../logging';
 
-/* eslint-disable-next-line import/no-commonjs, @typescript-eslint/no-var-requires */
```
boesing, line 5: I am pretty sure that I had to do this.
But since tests are passing, I think its okay now.

Maybe that was on my macos machine when I compiled that locally. Dunno...

Ocramius (approved):
LGTM, althought to be noted that XMLLint could (in theory) rely on the pre-defined XSD references within `phpunit.xml(.dist)`

Issue #22, opened 2021-03-21
Add composer.json linting to every matrix


Slamdunk:
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | no
| BC Break    | no

#### Summary

Hi, I'd like to add to every matrix one with the purpose of linting `composer.json`. In [Paratest](https://github.com/paratestphp/paratest/blob/v6.2.0/.github/workflows/integrate.yaml#L12) we use 4 steps:

| Command | Reference |
| --- | --- |
| `composer validate --strict` | https://getcomposer.org/doc/03-cli.md#validate |
| `composer-normalize --dry-run` | https://github.com/ergebnis/composer-normalize |
| `composer-require-checker check` | https://github.com/maglnet/ComposerRequireChecker |
| `composer-unused` | https://github.com/composer-unused/composer-unused |

weierophinney:
I'd argue we should do validation directly in this action, and then have the action generate one or more jobs that performs the require-checker and unused checks (normalize is really only interesting if we do a diff and/or autocommit the changes).

(Validation in this action would be required, as it would guarantee that the `composer install/update` operations in the CI action do not fail.)

Slamdunk:
> I'd argue we should do validation directly in this action
> Validation in this action would be required, as it would guarantee that the `composer install/update` operations in the CI action do not fail.

That'd be great, but how would you do that, considering that [`composer validate`](https://github.com/composer/composer/blob/master/src/Composer/Command/ValidateCommand.php) command needs PHP runtime, but this actions runs on `node`?

> normalize is really only interesting if we do a diff and/or autocommit the changes

That is right, but I don't think it's worth the effort of building such tool.
After the initial normalization, for which I can open a PR on all laminas packages with a trivial script to run on my pc, `composer normalize` rarely fails because who edits the `composer.json` easily adapt the new changes to the present style.

weierophinney:
> That'd be great, but how would you do that, considering that composer validate command needs PHP runtime, but this actions runs on node?

Composer [publishes its schema](https://getcomposer.org/schema.json), and we can use node-based tooling to validate the `composer.json` file against it.

boesing:
I've added a schema valiation in #33 

#### composer validate --strict
After this, I think it could be worth adding `composer validate --strict --no-check-all` to keep things simple.

#### composer-normalize
`composer-normalize` will probably fail in some projects, so I'd say we should have this as opt-in.

[...]

PR #32, opened 2021-05-17
Add composer-normalize

ghostwriter:
|    Q          |   A
|-------------- | ------
| QA            | yes

This PR add a QA Job to run `$ composer normalize --dry-run`.

If `composer.json` is not normalized (or `composer.lock` is not up-to-date), the command will fail with an exit code of `1` and show a diff.

blocked by laminas/laminas-continuous-integration-action#38

### DIFF
--- a/src/create-jobs.js
+++ b/src/create-jobs.js
@@ -85,6 +85,17 @@ function checks (config) {
     return [
         new Check(
             config.code_checks,
+            [fileTest('composer.json')],
+            /**
+             * @param {Config} config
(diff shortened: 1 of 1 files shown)


boesing:
I've written something about this feature here:
https://github.com/laminas/laminas-ci-matrix-action/issues/22#issuecomment-878601325

Not sure if we should add it as a failing check to the matrix tho. But in the end, @weierophinney has the last word. Just my 2cents.

PR #33, opened 2021-07-12
feature: introduce `composer.json` schema validation

### LINKED ISSUE #22: Add composer.json linting to every matrix

boesing:
|    Q          |   A
|-------------- | ------
| BC Break      | no
| New Feature   | yes

### Description

This will verify that the `composer.json` of a project always matches the schema.
There are two caveats:
- Only verify `composer.json` if there is one within the project
- Copied `composer.schema.json` to this project to ensure it wont change

The reason why I've added the schema to this project is, that the schema might change in the future.
Lets take `PSR-0` for example. There is absolutely no reason to support that for like forever.
As the `schema.json` on the getcomposer.org website is not versionized, I wanted to take this precaution.

I am open to change that to the https://getcomposer.org/schema.json if you think that we don't need this precaution.

This PR relates to #22 but only implements schema validation as suggested by @weierophinney

### DIFF
--- a/composer.schema.json
+++ b/composer.schema.json
@@ -1,0 +1,1288 @@
+{
+    "$schema": "https://json-schema.org/draft-04/schema#",
+    "title": "Package",
+    "type": "object",
+    "properties": {
+        "name": {
+            "type": "string",
+            "description": "Package name, including 'vendor-name/' prefix.",
+            "pattern": "^[a-z0-9]([_.-]?[a-z0-9]+)*/[a-z0-9](([_.]?|-{0,2})[a-z0-9]+)*$"
+        },
+        "type": {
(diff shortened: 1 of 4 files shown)


weierophinney (approved):
:ship:

PR #34, opened 2021-08-06
Codeception checks 

boesing:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| New Feature   | yes

### Description

Adds Codeception checks based on the presence of `codeception.yml.dist` or `codeception.yml` files.


weierophinney (approved):
:ship:

I know we don't necessarily use this for Laminas, but I also know others are adopting our CI, so it makes sense to accommodate it!

boesing:
I am not sure if this might use all php versions as well with minor/latest - so the same way as we do for unit testing now.
but we can optimize this later on as well.

PR #35, opened 2021-08-06
Execute phpunit with locked version only on the minimum supported PHP version of a project

boesing:
|    Q          |   A
|-------------- | ------
| Documentation | no
| BC Break      | yes
| New Feature   | yes

### Description

As discussed in the latest TSC meeting on [2021-08-02](https://github.com/laminas/technical-steering-committee/blob/c7f6babfb1773d9cb9371208102cbb8b27980bd5/meetings/minutes/2021-08-02-TSC-Minutes.md), this changes the way on how PHP unit tests are being executed.

`locked` dependencies are used for PHPUnit with the minimum supported PHP version of a package.
This also changes the way QA checks are being executed. All QA checks are now executed on the minimum supported PHP version of a project as well.

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -3,6 +3,8 @@ import semver from 'semver';
 import { Requirements } from './check-requirements.js';
 
 const CURRENT_STABLE       = '7.4';
+
+/** NOTE: Please keep this list ordered as the ordering is used to detect the minimum supported version of a project */
 const INSTALLABLE_VERSIONS = [
     '5.6',
     '7.0',
@@ -60,6 +62,8 @@ class Config {
     doc_linting            = true;
     versions               = [];
     stable_version         = CURRENT_STABLE;
+    minimum_version        = CURRENT_STABLE;
+    locked_dependencies    = false;
     extensions             = [];
     php_ini                = ['memory_limit        = -1'];
     dependencies           = ['lowest', 'latest'];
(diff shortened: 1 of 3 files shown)


src/config.js
```diff
@@ -81,6 +85,11 @@ class Config {
 
         if (configuration["stablePHP"] !== undefined) {
             this.stable_version = configuration["stablePHP"];
+            this.minimum_version = this.stable_version;
+        }
+
+        if (this.versions.length > 0) {
+            this.minimum_version = this.versions[0]
```
weierophinney, line 92: Are you sure they're sorted when you do this?

weierophinney (commented):
Looks reasonable; main concern is that you're actually getting the minimum version, and not a random one.

src/config.js
boesing: since we are iterating over the list of `INSTALLABLE_VERSIONS`, I think yes. But I did not try that, but can do so.

boesing:
I did some checks and it absolutely works:

```
→ cat composer.json 
{
    "require": {
        "php": "~8.0.0 || ^7.4"
    }
}

→ node index.js | grep "Using minimum"
Using minimum PHP version: 7.4
```

```
→ cat composer.json 
{
    "require": {
        "php": "^5.6 || 7.3 || ^8.0"
    }
}

→ node index.js | grep "Using minimum"
Using minimum PHP version: 5.6
```

```
→ cat composer.json 
{
    "require": {
        "php": "^5.3 || ^7.0 || ^8.0"
    }
}

→ node index.js | grep "Using minimum"
Using minimum PHP version: 5.6
```

It is as of now: yes, a project may state to support PHP <5.6 but CI will only run ^5.6 as we do not have other versions installed tho.

weierophinney (approved):
:ship:

weierophinney:
Removed the BC break label; it's a change to the generated matrix, but it's an _improvement_, as it provides better clarity for which version the lockfile should target, as well as reduces the number of jobs necessary to accomplish the same goal (determine if a new failure is due to adopting new features not supported in a minimum version of a requirement, or due to a BC breaking change in a later version of the requirement).

PR #46, opened 2021-08-17
Ignore platform requirements object

boesing:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| BC Break      | no
| New Feature   | yes

### Description

With one of the earlier versions, we introduced `ignore_platform_reqs_8` to ignore PHP 8.0 composer dependency conflicts.
Since we will need this for PHP 8.1 as well, this PR adds a new way to ignore platform requirements for **any** PHP version.

New format now is:

```json
{
    "ignore_php_platform_requirements": {
        "7.4": true,
        "8.0": false,
        "8.1": true   
    }
}
```

PHP 8.0 and PHP 8.1 are enabled by default or do we want to change default behavior for PHP 8.0 with the next release?

This PR also deprecates the usage of `ignore_platform_reqs_8` in projects and will add a warning to the actions output.

### DIFF
--- a/README.md
+++ b/README.md
@@ -171,7 +171,7 @@ The syntax for the `additional_checks` key is as follows:
 
 A job per PHP version per dependency set will be created, and the "name" will be appended with "on PHP {VERSION} with {DEPS} dependencies" during an actual run.
 The "job" element MUST have the structure as shown [here](#job-element) **but** the "php" element is mandatory in here and MUST contain either the minor PHP version to run the check against or a wildcard `*` to run the against **all** supported PHP versions.
-In case that the "php" version is passed as a wildcard `*`, the `ignore_php_platform_requirement` element will be ignored. It is possible to provide per-version flags by adding the `ignore_php_platform_requirements` element instead.
+You can pass the wildcard, "*", for the "php" element; when you do, the `ignore_php_platform_requirement` element will be ignored. It is possible to provide per-version flags by adding the `ignore_php_platform_requirements` element instead.
 
 The tool discovers checks first, then appends any `additional_checks` are concatenated, and then any `exclude` rules are applied.
 
--- a/README.md
+++ b/README.md
@@ -92,8 +92,9 @@ The "job" element will have the following elements, but is not restricted to the
     "(optional) php.ini directives, one per element; e.g. 'memory_limit=-1'"
   ],
   "dependencies": "(optional) dependencies to test against; one of lowest, locked, latest. default: locked",
-  "command": "(required) command to run to perform the check",
-  "ignore_platform_reqs_8": "(optional) boolean; whether to add `--ignore-platform-req=php` to composer for PHP 8.0. default: true"
+  "command": "(required) command to run to perform the check", 
+  "ignore_platform_reqs_8": "(optional; deprecated) boolean; whether to add `--ignore-platform-req=php` to composer for PHP 8.0. default: true",
+  "ignore_php_platform_requirement": "(optional) boolean; whether to add `--ignore-platform-req=php` to composer for this job."
 }
 ```
 
@@ -121,7 +122,9 @@ The package can include a configuration file in its root, `.laminas-ci.json`, wh
     {
     }
   ],
-  "ignore_platform_reqs_8": true,
+  "ignore_php_platform_requirements": {
+      "8.0": true
+  },
   "stablePHP": "7.4"
 }
 ```
@@ -168,6 +171,7 @@ The syntax for the `additional_checks` key is as follows:
 
 A job per PHP version per dependency set will be created, and the "name" will be appended with "on PHP {VERSION} with {DEPS} dependencies" during an actual run.
 The "job" element MUST have the structure as shown [here](#job-element) **but** the "php" element is mandatory in here and MUST contain either the minor PHP version to run the check against or a wildcard `*` to run the against **all** supported PHP versions.
+In case that the "php" version is passed as a wildcard `*`, the `ignore_php_platform_requirement` element will be ignored. It is possible to provide per-version flags by adding the `ignore_php_platform_requirements` element instead.
 
 The tool discovers checks first, then appends any `additional_checks` are concatenated, and then any `exclude` rules are applied.
 
--- a/index.js
+++ b/index.js
@@ -20,10 +20,7 @@ if (fs.existsSync('composer.json') && fs.existsSync('/action/composer.schema.jso
 
     if (!validationResult.valid) {
         validationResult.errors.forEach(function (outputUnit) {
-           core.error("There is an error in the keyword located by {0}: {1}".format(
-               outputUnit.keywordLocation,
-               outputUnit.error
-           ));
+           core.error(`There is an error in the keyword located by ${outputUnit.keywordLocation}: ${outputUnit.error}`);
         });
         core.setFailed('composer.json schema validation failed');
         process.exit(1);
--- a/index.js
+++ b/index.js
@@ -10,7 +10,7 @@ import {Validator} from "@cfworker/json-schema";
  * Do early composer.json schema validation to avoid unnecessary ramp-ups of jobs which may fail
  * due to an incompatible composer.json.
  */
-if (fs.existsSync('composer.json')) {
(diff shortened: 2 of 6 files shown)


Ocramius:
> PHP 8.0 and PHP 8.1 are enabled by default 

@_@ what

boesing:
Keeping BC 🤷🏼‍♂️ 
PHP 8.0 is still ignored platform reqs by default.
I am open to disable this and let projects enable that by themselves when adding PHP 8.1 support 👍🏼

Ocramius:
If we started supporting `~8.0.0`, we shouldn't have run with `--ignore-platform-reqs` on it :-\

boesing:
Imho, having `--ignore-platform-req=php` for PHP 8.0 was the only reason why we were able to provide support for PHP 8.0 early on.
We have 160 packages which depend on probably even more packages (even tho, most of them are part of laminas) but if we could only provide PHP 8.0 support if the whole dependency tree made its changes to be compatible, we will end up waiting for months before we can start. Thus said, ignoring platform reqs was decent for providing initial support for PHP 8.0.

boesing:
This will now output the following matrix when no further `.laminas-ci.json` is provided:

```
{
    "include":
    [
        {
            "name": "PHPUnit on PHP 7.3 with lowest dependencies",
            "job": "{\"command\":\"./vendor/bin/phpunit\",\"php\":\"7.3\",\"extensions\":
            [],\"ini\":[\"memory_limit        = -1\"],\"dependencies\":\"lowest\",\"ignore_platform_reqs_8\":false,\"ignore_php_platform_requirement\":false}",
            "operatingSystem": "ubuntu-latest",
            "action": "laminas/laminas-continuous-integration-action@v1"
        },
        {
            "name": "PHPUnit on PHP 7.3 with latest dependencies",
            "job": "{\"command\":\"./vendor/bin/phpunit\",\"php\":\"7.3\",\"extensions\":
            [],\"ini\":[\"memory_limit        = -1\"],\"dependencies\":\"latest\",\"ignore_platform_reqs_8\":false,\"ignore_php_platform_requirement\":false}",
            "operatingSystem": "ubuntu-latest",
            "action": "laminas/laminas-continuous-integration-action@v1"
        },
        {
            "name": "PHPUnit on PHP 7.4 with lowest dependencies",
            "job": "{\"command\":\"./vendor/bin/phpunit\",\"php\":\"7.4\",\"extensions\":
            [],\"ini\":[\"memory_limit        = -1\"],\"dependencies\":\"lowest\",\"ignore_platform_reqs_8\":false,\"ignore_php_platform_requirement\":false}",
            "operatingSystem": "ubuntu-latest",
            "action": "laminas/laminas-continuous-integration-action@v1"
        },
        {
            "name": "PHPUnit on PHP 7.4 with latest dependencies",
            "job": "{\"command\":\"./vendor/bin/phpunit\",\"php\":\"7.4\",\"extensions\":
            [],\"ini\":[\"memory_limit        = -1\"],\"dependencies\":\"latest\",\"ignore_platform_reqs_8\":false,\"ignore_php_platform_requirement\":false}",
            "operatingSystem": "ubuntu-latest",
            "action": "laminas/laminas-continuous-integration-action@v1"
        },
        {
            "name": "PHPUnit on PHP 8.0 with lowest dependencies",
            "job": "{\"command\":\"./vendor/bin/phpunit\",\"php\":\"8.0\",\"extensions\":
            [],\"ini\":[\"memory_limit        = -1\"],\"dependencies\":\"lowest\",\"ignore_platform_reqs_8\":true,\"ignore_php_platform_requirement\":true}",
            "operatingSystem": "ubuntu-latest",
            "action": "laminas/laminas-continuous-integration-action@v1"
        },
        {
            "name": "PHPUnit on PHP 8.0 with latest dependencies",
            "job": "{\"command\":\"./vendor/bin/phpunit\",\"php\":\"8.0\",\"extensions\":
            [],\"ini\":[\"memory_limit        = -1\"],\"dependencies\":\"latest\",\"ignore_platform_reqs_8\":true,\"ignore_php_platform_requirement\":true}",
            "operatingSystem": "ubuntu-latest",
            "action": "laminas/laminas-continuous-integration-action@v1"
        }
    ]
}
```

So while keeping BC compatibility by providing `ignore_platform_reqs_8` (when the PHP version for the job is PHP 8.0), we also add `ignore_php_platform_requirement` for the very specific PHP version which is used to execute the Job.

If we want to merge this, I will prepare a PR within the CI action afterwards.

src/command.js
```diff
@@ -32,7 +32,8 @@ export class Command {
             extensions: this.extensions,
             ini: this.ini,
             dependencies: this.dependencies,
-            ignore_platform_reqs_8: this.ignore_platform_reqs_8,
+            ignore_platform_reqs_8: this.ignore_php_platform_requirement,
```
boesing, line 35: When running a job on PHP 7.4, it is absolutely fine to provide `ignore_platform_reqs8` as `false` or `true` depending on the PHP 7.4 configuration.

The action itself will only apply this to the composer command if the PHP version used in the action is PHP 8.0 and thus, we are good to go.

boesing:
This PR contains changes from #47

src/command.js
weierophinney: This is good - it will ensure that the CI container doesn't need immediate changes, and we can update it _after_ this is already released.

weierophinney (commented):
Overall, this is good. I'd like to get #47 merged first, so that the diff on the `README.md` file is less, before merging this one.

As to the comments by @Ocramius - this is about installing dependencies. There's a fair number of dependencies of ours that do not have their PHP constraints updated, though they work fine with it. By applying the `--ignore-platform-req` option, we can still test against them.

That said, we should likely start seeing if we can test without that flag as we start the PHP 8.1 rollout, though it will be tempered by needing it for 8.1 tests ... which this patch would allow for.

src/command.js
boesing: > This is good - it will ensure that the CI container doesn't need immediate changes, and we can update it after this is already released.

Yup, that was the idea. This change is fully BC compatible.

index.js
```diff
@@ -10,7 +10,7 @@ import {Validator} from "@cfworker/json-schema";
  * Do early composer.json schema validation to avoid unnecessary ramp-ups of jobs which may fail
  * due to an incompatible composer.json.
  */
-if (fs.existsSync('composer.json')) {
+if (fs.existsSync('composer.json') && fs.existsSync('/action/composer.schema.json')) {
```
boesing, line 13: This is kinda unrelated but it always bit me when running this locally due to the absolute path requirement.
I can drop this again.

README.md
```diff
@@ -168,6 +171,7 @@ The syntax for the `additional_checks` key is as follows:
 
 A job per PHP version per dependency set will be created, and the "name" will be appended with "on PHP {VERSION} with {DEPS} dependencies" during an actual run.
 The "job" element MUST have the structure as shown [here](#job-element) **but** the "php" element is mandatory in here and MUST contain either the minor PHP version to run the check against or a wildcard `*` to run the against **all** supported PHP versions.
+In case that the "php" version is passed as a wildcard `*`, the `ignore_php_platform_requirement` element will be ignored. It is possible to provide per-version flags by adding the `ignore_php_platform_requirements` element instead.
```
weierophinney:
```suggestion
You can pass the wildcard, "*", for the "php" element; when you do, the `ignore_php_platform_requirement` element will be ignored. It is possible to provide per-version flags by adding the `ignore_php_platform_requirements` element instead.
```

index.js
```diff
@@ -20,10 +20,7 @@ if (fs.existsSync('composer.json')) {
 
     if (!validationResult.valid) {
         validationResult.errors.forEach(function (outputUnit) {
-           core.error("There is an error in the keyword located by {0}: {1}".format(
```
boesing, line 23: This is also deprecated but when I tried to add https://github.com/laminas/laminas-ci-matrix-action/pull/46/files#diff-e727e4bdf3657fd1d798edcd6b099d6e092f8573cba266154583a746bba0f346R48-R50, I had issues with `String.format` and thus, I wanted to ensure that this will definitely work.

weierophinney (changes requested):
One nit-pick, and thien this is ready.

weierophinney (approved):

PR #47, opened 2021-08-17
Update README to reflect actual features

boesing:
|    Q          |   A
|-------------- | ------
| Documentation | yes

### Description

While working on implementing #46, I realized that the documentation does not fully reflect any feature which is hidden when having a deep-dive into the code.

I detected these missing features:
- `additional_checks` "job" element can provide either `{"php": "7.4"}` or `{"php": "*"}`, the latter will execute that job on every PHP version supported by the project
- non-empty `checks` will prevent any auto-detection of jobs
- `ignore_platform_reqs_8` was missing in the `.laminas-ci.json` example
- `stablePHP` was missing in the `.laminas-ci.json` example

Furthermore, the documentation repeated the `job` configuration multiple time while not being in sync.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,20 +76,24 @@ The "exclude" element will only be present if the package using the action provi
 Each item in the "exclude" array will be an object, with one or more of the keys listed in the "include" objects; when a job matches all elements specified in the "exclude" array, it will be excluded from runs.
 
 The "job" element is a string JSON object detailing the job to run.
-Note: it is **not** an object; it is a JSON string.
-It will have the following elements, but is not restricted to them:
+Note: it is **not** an object; it is a JSON string but it MUST have the structure as shown [here](#job-element).
+
+### Job Element
+
+The "job" element will have the following elements, but is not restricted to them:
 
 ```json
 {
-  "php": "string PHP minor version to run against",
+  "php": "(optional) string PHP minor version to run against",
   "extensions": [
-    "extension names to install; names are from the ondrej PHP repository, minus the php{VERSION}- prefix",
+    "(optional) extension names to install; names are from the ondrej PHP repository, minus the php{VERSION}- prefix"
   ],
   "ini": [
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -76,20 +76,24 @@ The "exclude" element will only be present if the package using the action provi
...
   ],
   "ini": [
-    "php.ini directives, one per element; e.g. 'memory_limit=-1'",
+    "(optional) php.ini directives, one per element; e.g. 'memory_limit=-1'"
   ],
-  "dependencies": "dependencies to test against; one of lowest, locked, latest",
-  "command": "command to run to perform the check"
+  "dependencies": "(optional) dependencies to test against; one of lowest, locked, latest. default: locked",
+  "command": "(required) command to run to perform the check",
+  "ignore_platform_reqs_8": "(optional) boolean; whether to add `--ignore-platform-req=php` to composer for PHP 8.0"
```
weierophinney: We should indicate the default value here (it's `true`).

README.md
```diff
@@ -130,34 +136,18 @@ Each element in that array should be in the same format as listed above for the
...
-  "ignore_platform_reqs_on_8": "(boolean; OPTIONAL) Whether or not to ignore platform requirements on PHP 8; defaults to true",
-  "command": "command to run to perform the check"
-}
-```
+In each case, it MUST have the structure as shown [here](#job-element) **but** the "php" element is mandatory in here and MUST contain
+the minor PHP version to run the check against.
 
 The action validates each check and its job to ensure it is structured correctly, and will provide warnings if not, omitting any check that is malformed from the output.
 
+**If specific checks are provided by a project, the matrix won't contain any auto-detected checks and will only output these checks.**
```
weierophinney:
```suggestion
**If the `checks` element is populated, the matrix will not include any auto-detected checks, and will only output the jobs listed in that element.**
```

weierophinney (changes requested):

weierophinney (approved):

PR #59, opened 2021-09-19
Add `infection/infection` to CI runs, when `infection.json(.dist)` exists

### LINKED ISSUE #3: Include infection and roave/infection-static-analysis-plugin to matrix
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | no
| BC Break    | no

#### Summary

As discussed in https://github.com/laminas/automatic-releases/pull/118, we want to include `infection` and `roave/infection-static-analysis-plugin` to matrix here - so that we can also make automatic-releases use our brand new CI matrix process.

The only thing we need to verify is when to add the `infection`, or `roave/infection-static-analysis-plugin`, or both to the build matrix.

Ocramius:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | yes
| RFC           | no
| QA            | no

### Description

Fixes #3

Adds `vendor/bin/infection` to executed CI pipeline matrix, when `infection.json` or `infection.json.dist` exist.

~~Note: I've not yet included `roave/infection-static-analysis-plugin`, which requires checking `composer.json` or `vendor/bin/roave-infection-static-analysis-plugin` (probably the former, since `composer install` happens inside the job, and not in the matrix)~~

I'm quite uncomfortable with this tooling not having tests/QA on its own, so I don't have a valid way to verify if my code does what it should do :thinking:

### DIFF
--- a/src/create-jobs.js
+++ b/src/create-jobs.js
@@ -6,6 +6,7 @@ import { Config } from './config.js';
 import { Job } from './job.js';
 import create_additional_jobs from './additional-checks.js';
 import validateAndNormalizeChecks from './validate-and-normalize-checks-from-config.js';
+import parseJsonFile from "./json.js";
 
 /**
  * @param {String} filename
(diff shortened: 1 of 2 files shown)


src/create-jobs.js
```diff
@@ -157,6 +157,48 @@ function checks (config) {
                 return createQaJobs('./vendor/bin/phpbench run --revs=2 --iterations=2 --report=aggregate', config);
             }
         ),
+        new Check(
+            config.code_checks,
+            [
+                function () {
+                    const composerFile = JSON.parse(fs.readFileSync('composer.json'));
```
boesing: I would prefer having this added to the first line of the `check` function so it can be reused in both `infection` and `roave-infection-static-analysis-plugin` check

boesing (changes requested):

boesing:
> I'm quite uncomfortable with this tooling not having tests/QA on its own, so I don't have a valid way to verify if my code does what it should do 🤔

Usually, I do add `composer.json` to the root directory of this component containing those things you want to verify and then execute `node index.js`.

This will output the JSON and you can verify if it works as intended.

src/create-jobs.js
```diff
@@ -157,6 +157,48 @@ function checks (config) {
...
+            }
+        ),
+        new Check(
+            config.code_checks,
+            [
+                function () {
+                    const composerFile = JSON.parse(fs.readFileSync('composer.json'));
+
+                    return composerFile.hasOwnProperty('require-dev') &&
+                        composerFile['require-dev'].hasOwnProperty('roave/infection-static-analysis-plugin');
```
boesing: so does this plugin won't need the `infection.json`?

boesing (changes requested):

src/create-jobs.js
Ocramius: It does, but a better/simpler check is just to see if the dependency is there

boesing: maybe checking both as for infection? you could check in infection and just swap the commands. So we end up having one new check rather than two which are mutually exclusive?

boesing:
@Ocramius I think I will start working on this on sunday as I want to kick out 1.11.0 as there are plenty of features I've added in the past weeks.

If you want to dive into this by yourself, just hit me up.

boesing:
@Ocramius would you mind double-checking the recent changes? If this suits your needs, I'd love to merge this PR

boesing:
Okay, I'll summarize my changes:

I've moved the `roave/infection-static-analysis-plugin` into the same check which also verifies the existence of `infection.json` or `infection.json.dist`.

The rationale for this is that infection cannot work without these files and thus, we can merge the initially created checks (one for both) into one check.
image

So in the end, we still fulfill the initial requirements (preferring the roave plugin over infection itself in case it is installed).

Ocramius:
@boesing interestingly, all my CI actions failed today :D 

Reason: xdebug or pcov missing.

What's the best way to handle that, in your opinion? Should a `require-dev` addition be added? Should the CI setup install pcov when running infection?

boesing:
As mentioned in slack: adding `xdebug` or `pcov` to `require-dev` or `.laminas-ci.json` would be fine.
But it seems that its fine to execute with `phpdbg` as you pointed out in #70 👍🏼

Ocramius:
I mostly preferred going with phpdbg because it is developed **together** with the PHP engine. Means 8.2 will have phpdbg out of the box :) 

XDebug and pcov are fine, but there's always some wait time in between

PR #62, opened 2021-09-24
feature: introduce CI

boesing:
|    Q          |   A
|-------------- | ------
| New Feature   | yes
| QA            | yes

### Description

Initial work on adding CI pipeline for the matrix action.


boesing:
I see, we need a slightly changed container so it does not always do the composer magic...

boesing:
@weierophinney Any chance that we get this merged?
I've added plenty of tests already and would be happy to take this as a base for #83 so we can finally ship that whole thing.

.github/workflows/continuous-integration.yml
```diff
@@ -0,0 +1,98 @@
...
+    name: Generate job matrix
+    runs-on: ubuntu-latest
+    outputs:
+      matrix: ${{ steps.matrix.outputs.matrix }}
+    steps:
+      - name: Checkout sourcecode
+        uses: actions/checkout@v3
+      - name: Gather integration test directory names
+        id: matrix
+        run: cd tests; echo "::set-output name=matrix::[\"$(ls -d * | tr '\n' ' ' | sed 's/ $//' | sed 's/ /\",\"/g')\"]"
```
Ocramius, line 26: Maybe a bit cryptic - TBH, a few comments in this file would clarify that we're running a bunch of parallel tests to check that the action will generate the right JSON for the various directories in here.

That said, this is 💯 

Awesome!

Ocramius (approved):
IMO 🚢

Possibly worth merging this one before #83, to see if we are at parity too.

.github/workflows/continuous-integration.yml
boesing: True. Will add this as a comment. Will merge this afterwards and then rebase #83

PR #70, opened 2021-10-11
Run `infection` via `phpdbg` by default, since `phpdbg` is pre-installed with PHP

Ocramius:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description
This removes the need for installing `pcov` or `xdebug` in environments that need to
run mutation tests: `phpdbg` being pre-installed means that any bugs with it will
also be reported to the PHP project, instead of dragging in bugs from third-party
extensions.

This also fixes builds for projects that did not declare `ext-pcov` or `ext-xdebug`
as dependencies in `require-dev`: no need to have those, when running this task.

### DIFF
--- a/src/create-jobs.js
+++ b/src/create-jobs.js
@@ -181,13 +181,12 @@ function checks (config) {
              */
             function (config) {
                 const composerFile = parseJsonFile('composer.json', false);
-                let commandToExecute = './vendor/bin/infection';
 
                 if (composerFile.hasOwnProperty('require-dev') && composerFile['require-dev'].hasOwnProperty('roave/infection-static-analysis-plugin')) {
-                    commandToExecute = './vendor/bin/roave-infection-static-analysis-plugin';
+                    return createQaJobs('phpdbg -qrr ./vendor/bin/roave-infection-static-analysis-plugin', config);
                 }
 
-                return createQaJobs(commandToExecute, config);
+                return createQaJobs('phpdbg -qrr ./vendor/bin/infection', config);
             }
         ),
         new Check(


boesing:
I am not sure if `phpdbg` is set by-version when running the container.

We do actually only link `php-config`, `php` and `phpize`.

https://github.com/laminas/laminas-continuous-integration-action/blob/a2e86382379d207e43296a442e16646d4b67f0da/entrypoint.sh#L136

Not sure what kind of binary `phpdbg` is so it might work but I doubt it.

Ocramius:
As discussed with @boesing, will need to adjust `laminas-continuous-integration-action` too - investigating/preparing a patch

boesing (approved):
LGTM

https://github.com/laminas/laminas-continuous-integration-action/releases/tag/1.13.2 is already published and thus this should work.

You can verify this by creating the matrix with:

```
docker build . -t laminas/laminas-ci-matrix-action:dev-local
# The next command should output the JSON to pass to the container
docker run -v $(realpath .):/github/workspace -w=/github/workspace laminas/laminas-ci-matrix-action:dev-local
# The next command should pull the latest version of the CI Container
docker pull ghcr.io/laminas/laminas-continuous-integration:1
# Pass the JSON to the container
docker run -v $(realpath .):/github/workspace -w=/github/workspace ghcr.io/laminas/laminas-continious-integration-action:1 '{}'

Ocramius:
@boesing just verified this on `ocramius/proxy-manager` and it works :+1: 

:ship: + :label:

PR #74, opened 2022-01-20
Normalize composer version range

### LINKED ISSUE #73: Missing jobs due to incompatibility composer PHP version parser
### Bug Report

|    Q        |   A
|------------ | ------
| Version(s)  | <=1.11.4

#### Summary

the semver version parser does not work correctly if `composer.json` uses this format `"php": ">=7.4, <8.2"`, which prevents certain jobs from being created.

https://github.com/laminas/laminas-ci-matrix-action/blob/71aa5bef47be3a9d3e1a3865f43caf16708ed847/src/config.js#L25-L39

#### How to reproduce

setup a repo with a `composer.json` file with `"require": {"php": ">=7.4, <8.2"}` .

#### Expected behavior

Generate jobs for PHPUnit on PHP `7.4`, `8.0`, & `8.1`

ghostwriter:
|    Q          |   A
|-------------- | ------
| Bugfix        | yes

This patch resolves #73.

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -28,9 +28,10 @@ function gatherVersions(composerJson) {
     }
 
     let versions = [];
+    let composerPhpVersion = composerJson['require']['php'].replace(/,\s/i, ' ');
 
     INSTALLABLE_VERSIONS.forEach(function (version) {
-        if (semver.satisfies(version + '.0', composerJson['require']['php'])) {
+        if (semver.satisfies(version + '.0', composerPhpVersion)) {
             versions.push(version);
         }
     });


Ocramius (approved):
IMO OK, but the fact that this repo has zero tests/quality checks (also acting as documentation) worries me quite a bit 😬

@weierophinney do we have a plan for this somewhere?

boesing:
@Ocramius I've started working on CI last year: https://github.com/laminas/laminas-ci-matrix-action/pull/62
The main idea was to create a directory where "projects" can be created with specific "files" and with a `matrix.json`. This would enable us to create fake-projects with the exact outcome we expect when these are "parsed" by this container.

But sadly, I haven't written down that so that some1 else could take over.

PR #77, opened 2022-03-03
Add `exclude_strict_matching` configuration

### LINKED ISSUE #76: Identify a check by name within CI action
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | yes
| BC Break    | no

#### Summary

Hey there,

I am currently working on something and I have the following use-case:
- run tests on all supported PHP versions (works out-of-the-box)
- run tests on lowest PHP version with locked deps (works ootb)
- run tests on lowest PHP version with locked deps via `additional_checks` (works ootb, **but**...)

So the latter does work due to prior changes made to this action. The problem is, that I now want to identify this exact `additional_check` somehow. Because the main idea of this check would be:

Run tests on lowest PHP version with locked deps **but** disable one or more PHP extensions (to see if tests are still passing due to precautions or polyfills) and/or disallow PHP functions.

So, I'd like to have some kind of optional `name` within the `job` configuration which can be used as some kind of unique identifier. When added, this can be extract within a `pre-run.sh` for example.
An alternative would be to actually allow disabling extensions via the `job` array. That would make the whole thing easier but maybe its worth having both "features" up and running.

WDYT?

boesing:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| Bugfix        | no
| BC Break      | no
| New Feature   | yes

### Description

Added a new feature named `exclude_strict_matching` which actually allows projects to exclude jobs based on partially matching values.

Examples:

```json
{
     "additional_checks": [
          {
               "name": "Do whatever",
               "job": {
                   "php": "*",
                   "dependencies": "*",
                   "command": "some command"
               }
          }
     ],
     "exclude": [
         {
             "name": "Do whatever on PHP 8."
         }
     ],
     "exclude_strict_matching": false,
}
```

In projects with multiple PHP versions supported, starting from PHP 7.4, every job on PHP 8.0+ will be excluded.

Prior to this change, one had to add multiple excludes with all combinations of PHP version + dependencies.

_Will add additional changes to address ideas posted by @Ocramius in https://github.com/laminas/laminas-ci-matrix-action/issues/76#issuecomment-1057552130_

### DIFF
--- a/laminas-ci.schema.json
+++ b/laminas-ci.schema.json
@@ -177,6 +177,9 @@
                 "exclude": {
                     "$ref": "#/definitions/exclude"
                 },
+                "exclude_strict_matching": {
+                    "$ref": "#/definitions/exclude_strict_matching"
+                },
                 "ignore_php_platform_requirements": {
                     "$ref": "#/definitions/ignore_php_platform_requirements"
                 },
@@ -309,6 +312,12 @@
                 },
                 "additionalProperties": false
             }
+        },
+        "exclude_strict_matching": {
+            "title": "Flag to toggle strict matching of exclude property values",
+            "type": "boolean",
+            "default": true,
+            "description": "When this flag is disabled, the property values for exclude do also check if the job values do start with the provided value. This becomes handy when one wants to exclude jobs based on their names. Every job gets the PHP version + composer dependency information appended and thus, exact matches are quite verbose."
         },
         "ignore_php_platform_requirements": {
             "type": "object",
--- a/src/create-jobs.js
+++ b/src/create-jobs.js
@@ -236,7 +236,7 @@ function checks (config) {
     ];
 }
 
-const excludeJob = function (job, exclusion) {
+const excludeJob = function (job, exclusion, strict_comparison) {
     let matches = 0;
     Object.keys(job).forEach(function (jobKey) {
         Object.keys(exclusion).forEach(function (excludeKey) {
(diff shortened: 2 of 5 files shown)


laminas-ci.schema.json
```diff
@@ -310,6 +313,12 @@
                 "additionalProperties": false
             }
         },
+        "exclude_strict_matching": {
+            "title": "Flag to toggle strict matching of exclude property values",
+            "type": "boolean",
+            "default": true,
+            "description": "When this flag is disabled, the property values for exclude do also check if the job values do start with the provided value. This becomes handy when one wants to exclude jobs based on their names. Every job gets the PHP version + composer dependency information appended and thus, exact matches are quite verbose."
+        },
```
froschdesign, line 321: I think this option should be mentioned in the readme file. Otherwise, this is only found in the schema.

https://github.com/laminas/laminas-ci-matrix-action#excluding-specific-jobs

boesing: Good catch! I forgot to add it there.

src/create-jobs.js
```diff
@@ -271,7 +282,7 @@ export default function (config) {
         .filter(function (job) {
             let keep = true;
             config.exclude.forEach(function (exclusion) {
-                keep = keep && ! excludeJob(job, exclusion);
+                keep = keep && ! excludeJob(job, exclusion, config.exclude_strict_matching);
```
internalsystemerror, line 285: I think you want a `!` on the `config.exclude_strict_matching` here as if set to true, it sets `strict_comparison` to true inside `excludeJob`, which is the opposite of the intent.

internalsystemerror (changes requested):

src/create-jobs.js
boesing: Have to double check this again but I did some integration test on my local machine and that actually worked as intended.
However, with #83 and #62, this will become pretty easy.

internalsystemerror: Looking at around line 254 of your version of this file, if the 3rd argument is true, then it will return early and won't allow loser matchings. Unless I've missed something (very possible) this appears to be the inverse of the intent.

boesing:
Closing this in favor of the idea of #93

PR #78, opened 2022-03-03
Remove redundant type checks

boesing:
|    Q          |   A
|-------------- | ------
| QA            | yes

### Description

Due to the fact that this action validates the projects `.laminas-ci.json` with the `laminas-ci.schema.json` file, we do not have to verify types in some codepaths anymore.

### DIFF
--- a/src/additional-checks.js
+++ b/src/additional-checks.js
@@ -8,24 +8,6 @@ import {Job} from "./job.js";
  * @return {Boolean}
  */
 const validateCheck = function (checkConfig) {
-    if (typeof checkConfig !== 'object' || checkConfig === null) {
-        // NOT AN OBJECT!
-        core.warning("Skipping additional check; not an object, or is null: " + JSON.stringify(checkConfig));
-        return false;
-    }
-
-    if (checkConfig.name === undefined || checkConfig.job === undefined) {
-        // Missing one or more required elements
-        core.warning("Skipping additional check due to missing name or job keys: " + JSON.stringify(checkConfig));
-        return false;
-    }
-
-    if (typeof checkConfig.job !== 'object' || checkConfig.job === null) {
-        // Job is malformed
-        core.warning("Invalid job provided for check; not an object, or is null: " + JSON.stringify(checkConfig.job));
-        return false;
-    }
-
     if (checkConfig.job.command === undefined) {
         // Job is missing a command
         core.warning("Invalid job provided for check; missing command property: " + JSON.stringify(checkConfig.job));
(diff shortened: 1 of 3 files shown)


src/additional-checks.js
```diff
@@ -8,24 +8,6 @@ import {Job} from "./job.js";
  * @return {Boolean}
  */
 const validateCheck = function (checkConfig) {
-    if (typeof checkConfig !== 'object' || checkConfig === null) {
```
Ocramius, line 11: We should really write some typescript :D

src/additional-checks.js
```diff
@@ -8,24 +8,6 @@ import {Job} from "./job.js";
...
-        // NOT AN OBJECT!
-        core.warning("Skipping additional check; not an object, or is null: " + JSON.stringify(checkConfig));
-        return false;
-    }
-
-    if (checkConfig.name === undefined || checkConfig.job === undefined) {
```
Ocramius, line 17: Shouldn't we validate this? Nothing in the type declaration above says that `checkConfig.job` is required

src/additional-checks.js
```diff
@@ -8,24 +8,6 @@ import {Job} from "./job.js";
...
-        // Missing one or more required elements
-        core.warning("Skipping additional check due to missing name or job keys: " + JSON.stringify(checkConfig));
-        return false;
-    }
-
-    if (typeof checkConfig.job !== 'object' || checkConfig.job === null) {
```
Ocramius, line 23: Shouldn't we validate this? Nothing in the type declaration above says that `checkConfig.job` must be an `object`

Ocramius (changes requested):

src/additional-checks.js
boesing: I am not that familiar with typescript yet.

Do you have examples on how to achieve all these optional things in typescript/javascript?
I guess if we have to convert this action to typescript, I rather keep the redundant checks as I am definitely no expert in this at all.

boesing:
@Ocramius Would `eslint` check would be suitable for this?

I could give this a try to provide at least a little CI pipeline for this repository. I've started in #62 but I wanted to use our current CI pipeline. Maybe that was the wrong way since testing a repository with the repository might lead to unexpected issues.

Could start adding `eslint` and then extend with integration tests which verify the return values for different project setups.

Ocramius:
`eslint` would certainly be a first step, yes.

Overall, this codebase is growing, and I'm concerned that it could grow in an uncontrolled way, which would make it hard to manage. Anything that helps us keeping complexity in check will help, and types are a must, if we don't have tests.

src/additional-checks.js
Ocramius: Changing to typescript here is obviously out of scope for this patch. As for the definition, I would expect something more precise than `{Object}` as input definition for `checkConfig`, but I don't know much about JS docblock syntax :thinking:

boesing:
#82 introduces ESLint and I am currently working on porting to typescript.
Closing this in favor of typescript.

PR #79, opened 2022-03-03
Provide proper job names for github UI

boesing:
|    Q          |   A
|-------------- | ------
| New Feature   | yes

### Description

This feature is based on the feedback from @ocramius in #76 and provides better namings to github UI.
This will also prevent #31 with its extended commands where we do check for linting before actually executing the command itself.

### DIFF
--- a/src/create-jobs.js
+++ b/src/create-jobs.js
@@ -22,13 +22,14 @@ const fileTest = function (filename) {
 };
 
 /**
+ * @param {String} name
  * @param {String} command
  * @param {Config} config
  * @return {Array}
  */
-const createQaJobs = function (command, config) {
+const createQaCheck = function (name, command, config) {
     return [new Job(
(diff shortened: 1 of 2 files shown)


src/create-jobs.js
```diff
@@ -244,7 +247,13 @@ const excludeJob = function (job, exclusion) {
                 return;
             }
 
-            if (job[jobKey] === exclusion[excludeKey]) {
+            let jobValue = job[jobKey],
+                exclusionValue = exclusion[excludeKey];
+
+            if (jobValue === exclusionValue) {
+                matches += 1;
+            } else if (excludeKey === 'name' && job.deprecatedName !== '' && exclusionValue === job.deprecatedName) {
```
boesing, line 255: This is actually the BC layer for the issue mentioned in https://github.com/laminas/laminas-ci-matrix-action/issues/76#issuecomment-1057746570

PR #82, opened 2022-03-16
Introduce ESLint

boesing:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | no
| QA            | yes

### Description

Adds ESLint to this project to provide at least some CI pipeline for the moment.


Dockerfile
```diff
@@ -14,7 +14,7 @@ ADD package.json /action/package.json
 ADD package-lock.json /action/package-lock.json
 ADD composer.schema.json /action/
 ADD laminas-ci.schema.json /action/
-RUN (cd /action ; npm install)
+RUN (cd /action ; npm install --no-save)
```
Ocramius: Potentially `npm ci`

package.json
```diff
@@ -1,8 +1,20 @@
 {
+  "engines": {
+    "node": ">=17.0.0"
```
Ocramius: Probably worth declaring `^` here

package.json
```diff
@@ -1,8 +1,20 @@
...
+  },
   "dependencies": {
     "@actions/core": "^1.2.6",
     "@cfworker/json-schema": "^1.8.2",
     "semver": "^7.3.4"
   },
-  "type": "module"
+  "type": "module",
+  "devDependencies": {
+    "@babel/eslint-parser": "^7.17.0",
```
Ocramius: Do we even need babel, for this to work? We're on a pretty new version of JS :thinking: 

Non-blocker, just a question.

src/additionalChecks.js
```diff
@@ -10,25 +10,29 @@ import {Job} from "./job.js";
...
-        core.warning("Skipping additional check; not an object, or is null: " + JSON.stringify(checkConfig));
+        core.warning(`Skipping additional check; not an object, or is null: ${  JSON.stringify(checkConfig)}`);
+
         return false;
     }
 
     if (checkConfig.name === undefined || checkConfig.job === undefined) {
         // Missing one or more required elements
-        core.warning("Skipping additional check due to missing name or job keys: " + JSON.stringify(checkConfig));
+        core.warning(`Skipping additional check due to missing name or job keys: ${  JSON.stringify(checkConfig)}`);
```
Ocramius: Spacing got a bit weird here :thinking: 

Applies to other `${` conversions too.

src/command.js
```diff
@@ -1,43 +1,58 @@
...
+
+    php = CURRENT_STABLE;
+
+    extensions = [];
+
+    ini = [];
+
+    dependencies = 'locked';
+
+    ignorePhpPlatformRequirement = false;
```
Ocramius, line 14: Aren't these changes potential BC breaks?

src/createJobs.js
```diff
@@ -27,49 +23,50 @@ const fileTest = function (filename) {
  * @return {Array}
  */
 const createQaJobs = function (command, config) {
-    return [new Job(
-        command + ' on PHP ' + config.minimum_version,
+    return [ new Job(
+        `${command  } on PHP ${  config.minimumVersion}`,
         JSON.stringify(new Command(
```
Ocramius, line 28: This `JSON.stringify()` will produce a data structure that is potentially quite different from what we had before :thinking:

Ocramius (changes requested):

package.json
boesing: Depends. Our container used alpine latest and thus I guess it always contains the very latest version of node. Is that captured by `^`? Not sure if the constraints are the same as for composer 🤷🏼‍♂️

package.json
boesing: yes, because the eslint default parser was not able to infer most of the code. I've tested that.

src/command.js
boesing: I'd say no as we do not provide an API here, right?
Only changing application code. Input configs and output matrix havent changed.

package.json
Ocramius: Yeah, those are SemVer, so IMO we should pin to a major

src/createJobs.js
boesing: why do you think so? cant really follow.
JSON.stringify is already part of the latest version?

src/additionalChecks.js
boesing: I've seen that. These were added by eslint `--fix`.
Gonna see if I can change this without eslint triggering errors.

Dockerfile
boesing: ah, is that an alias? TBH I usually do not use npm so gonna change that.

package.json
boesing: Do you have an idea on how to keep `package.json` and `Dockerfile` in sync then?
Because the main idea here was to always ensure that we do run on latest but with a minimum of `17.0.0`.
That `engines` config is parsed by `eslint` and thats why I've added it.

Is this something we could use renovate for?

package.json
boesing: Oh interesting, I have changed the config lately and maybe thats why babel is not required anymore.
I'll drop it as `eslint` actually also works without it. 👍🏼

package.json
boesing: Changed to `^17` along with using `node:17-alpine` in the `Dockerfile`.

src/createJobs.js
Ocramius: Sorry, I mean that the structure passed to it will be a bit different :)

src/command.js
Ocramius: Isn't this serialized and given to the next action?

boesing:
Currently working on switching to typescript while also restructuring the whole application.
Making some progress there but still not finished - I am not too familiar with typescript but I tried to get some feedback from known typescript devs.
Main idea would be to get rid of all the annotation magic and getting better overview of what is happening.

Will see if I can finish #62 with some integration tests of "projects" so we have a bunch of tests which can also be executed by the new typescript version of this application.

If that is something we are fine with, I'd close this PR to instead use `tslint` in the `typescript` version.
Cool?

src/command.js
boesing: Nope, there is a dedicated method `toJSON` for that.

src/createJobs.js
boesing: Ah I see what you mean and yes indeed, it changes a little bit.
Good catch. We are missing to call `toJSON` here. 

There is some "normalization" stuff going on in another function which converts the string back to an object and verifies properties.
There it will not properly match anymore. 👍🏼

Issue #54, opened 2021-09-12
Simplify `.laminas-ci.json`


boesing:
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | yes
| BC Break    | yes

#### Summary

The `.laminas-ci.json` allows to pass `checks` which overrides the `additional_checks` and thus are mutually exclusive. This makes the JSON schema more bloated as we have to provide two configuration combinations with either `checks` or `additional_checks`.

The `.checks[].job` element has is allowed to be a string which also makes it more error-prone.
It also states that it is possible to select an additional "action" which can be used with the CI matrix which is actually not really true as if someone choses another "action", that action also have to parse the `job` JSON-string and from this perspective, I dont really see the benefit of having `checks`.

Due to the fact that `.checks[].job` is a string, it also cannot be validated by the JSON schema which is feature-requested in #49.

Is there any project around which uses the `.checks` functionality from `.laminas-ci.json`? If not, I would prefer getting rid of that to simplify the matrix action.

boesing:
Will get easier with ~#83~ #95 in case we implement JSON schema generation from typescript interfaces.

PR #83, opened 2022-03-25
Refactoring: typescript

### LINKED ISSUE #80: Empty or non-existent `ignore_php_platform_requirements` does not disable PHP 8.0 default behavior
### Bug Report

|    Q        |   A
|------------ | ------
| Version(s)  | 1.11.5

#### Summary

When I've impelemented the `ignore_php_platform_requirements` feature, the main idea was to get rid of the old and yet deprecated `ignore_platform_reqs_8` feature, which defaulted to `true`.

I was assuming (and thought I've implemented it that way 😅) that when adding an empty `ignore_php_platform_requirements` object to the `.laminas-ci.json`, we do not explicitly mark PHP 8.0 as `true`.
Sadly, it does and thus, I'd like to either change this and/or remove `ignore_platform_reqs_8` entirely.

`ignore_php_platform_requirements` option is available since [v1.9.0](https://github.com/laminas/laminas-ci-matrix-action/releases/tag/1.9.0) (released in 08/2021)

#### Current behavior

Adding an empty `ignore_php_platform_requirements` does not prevent ignoring platform reqs for PHP 8.0.

#### How to reproduce

Add an empty `ignore_php_platform_requirements` to the `.laminas-ci.json` in a project with PHP 8.0 support and check matrix output. It contains `ignore_php_platform_requirement: true` for PHP 8.0.

#### Expected behavior

`ignore_php_platform_requirement` should be `false` for PHP 8.0.

### LINKED ISSUE #84: `@lowest` and `@latest` is not respected when project provides explicit `checks`
### Bug Report

|    Q        |   A
|------------ | ------
| Version(s)  | 1.11.4

#### Summary

I am currently working on providing integration tests in #62 
There, I've seen that `@lowest` and `@latest` is not properly handled by the matrix action when it comes to explicit checks provided by the projects `.laminas-ci.json`

#### Current behavior

When `@lowest` and/or `@latest` is used for `job.php`, the action does skip that job:
```
::warning::Invalid job provided; no PHP version or unknown PHP version specified: {"php":"@lowest","dependencies":"locked","command":"vendor/bin/phpunit"}
::warning::Invalid job provided; no PHP version or unknown PHP version specified: {"php":"@latest","dependencies":"locked","command":"vendor/bin/phpunit"}
```

##### `.laminas-ci.json`
```json
{
    "checks": [
        {
            "name": "PHPUnit with locked dependencies on lowest supported PHP version",
            "job": {
                "php": "@lowest",
                "dependencies": "locked",
                "command": "vendor/bin/phpunit"
            }
        },
        {
            "name": "PHPUnit with locked dependencies on lowest supported PHP version",
            "job": {
                "php": "@latest",
                "dependencies": "locked",
                "command": "vendor/bin/phpunit"
            }
        }
    ]
}
```

#### Expected behavior

Matrix does not report a warning for `@lowest` or `@latest` and properly uses the lowest and latest supported PHP version by the project within the job declaration.

boesing:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | no
| QA            | yes

### Description

Converted the application to typescript. 
In combination with #62, this supposed to be a major improvement when it comes to new features.

This also reduces the container size from **426MB** to **188MB** which is a total amount of **~56%**!

#### Features which are implemented in this PR
- #79 
- #82

#### Bugs which are fixed in this PR
- #84 
- #80 

#### Open TODOs
- [x] Implement BC layer for `exclude` which was implemented [here](https://github.com/laminas/laminas-ci-matrix-action/pull/79#discussion_r818499949)
- [x] Add PHP version to the check name
- [x] It seems that the `.laminas-ci.schema.json` is not properly verified, it does not contain `@latest` or `@lowest` for the `php` entry but integration tests are passing

#### Ideas which could be implemented later on
- [x] Generation of `.laminas-ci.schema.json` from typescript interfaces using [typescript-json-schema](https://www.npmjs.com/package/typescript-json-schema) - would target #54 (addressed in https://github.com/laminas/laminas-ci-matrix-action/issues/95)
- [x] Enable code checks or linters when files from tools are changed (i.e. run all PHPUnit checks when `phpunit.xml.dist` is the only file within the diff) (addressed in https://github.com/laminas/laminas-ci-matrix-action/issues/94)
- [x] Extend `exclude` to use the tools `name` along with a `php` and `dependencies` property: `{"name": "PHPUnit", "php": "*"}` (addressed in https://github.com/laminas/laminas-ci-matrix-action/issues/93)
- [x] Optimize `diff` output for better readability in case of differences (addressed in https://github.com/laminas/laminas-ci-matrix-action/issues/92)


.github/workflows/continuous-integration.yml
```diff
@@ -0,0 +1,22 @@
+name: "Continuous Integration"
+
+on:
+  pull_request:
+  push:
+    branches:
+      - '[0-9]+.[0-9]+.x'
+      - 'refs/pull/*'
```
internalsystemerror, line 8: If you're adding in `package.json`, do you also want to allow renovate patches here?

Dockerfile
```diff
@@ -8,13 +8,18 @@ RUN apk update \
...
-RUN (cd /action ; npm install)
+
+RUN mkdir -p /usr/local/source
+COPY . /usr/local/source
+ADD package.json /usr/local/source/package.json
+ADD package-lock.json /usr/local/source/package-lock.json
+ADD tsconfig.json /usr/local/source/tsconfig.json
+
+RUN (cd /usr/local/source; npm ci && npm run build)
+RUN mv /usr/local/source/dist/main.js /action/
```
internalsystemerror: Would it be better to use 2 different layers, one for the build, and the other for the runtime? I haven't thought too much about it, but my instincts are telling me it would be better for caching. Let me know if I'm wrong here.

composer.schema.json
```diff
@@ -3,1286 +3,20 @@
     "title": "Package",
     "type": "object",
     "properties": {
```
internalsystemerror: This file could be obtained from https://getcomposer.org/schema.json, or for versioning purposes, from https://github.com/composer/composer/blob/2.2/res/composer-schema.json.

package.json
```diff
@@ -1,8 +1,29 @@
 {
-  "dependencies": {
+  "name": "laminas-ci-matrix-action",
+  "engines": {
+    "node": "^17"
+  },
+  "type": "commonjs",
+  "devDependencies": {
```
internalsystemerror, line 7: There's no explicit dependency on typescript, and its not semver compliant.

src/action/github.ts
```diff
@@ -0,0 +1,30 @@
...
+export class Github implements Action {
+    debug(message: string): void {
+        core.debug(message)
+    }
+
+    info(message: string): void {
+        core.info(message)
+    }
+
+    error(message: string): void {
```
internalsystemerror: the names of these methods could be more descriptive, e.g. logError. It makes sense for the Logger interface, but when mixed with the action, at first glance one may assume `error` triggers the action to error or `debug` enables debug mode for the action, or something similar.

src/app.ts
```diff
@@ -0,0 +1,175 @@
...
+        let schemaValidationResult = this.projectContainsValidJsonConfiguration(
+            'composer.json.schema',
+            this.composerJsonFileName
+        );
+
+        if (!schemaValidationResult.valid) {
+            return this.failWithLoggedJsonSchemaValidationErrors(this.composerJsonFileName, schemaValidationResult);
+        }
+
+        schemaValidationResult = this.projectContainsValidJsonConfiguration(
```
internalsystemerror, line 34: Not that it really matters here, but you could use 2 const's over a let which I believe would be a micro-optimisation. (again, correct me if I'm wrong, this could be old news now)

src/app.ts
```diff
@@ -0,0 +1,175 @@
...
+        const config = parseJsonFile(this.continousIntegrationConfigurationJsonFileName, true) as ConfigurationFromFile;
+        const tools = createTools(appConfig);
+
+        this.action.debug(`Tools detected: ${JSON.stringify(tools, null, SPACES_TO_INDENT_JSON)}`);
+        const jobs: [Job, ...Job[]] = (new DefaultJobCreator(tools))
+            .createJobs(config, appConfig);
+
+        this.action.debug(`Jobs created: ${JSON.stringify(jobs, null, SPACES_TO_INDENT_JSON)}`);
+
+        return jobs.map((job) => createJobForMatrixFromJob(job)) as [JobForMatrix, ...JobForMatrix[]];
```
internalsystemerror, line 76: Just a question as to `[JobForMatrix, ...JobForMatrix[]]`, wouldn't this be the same as `JobForMatrix[]`?

src/app.ts
```diff
@@ -0,0 +1,175 @@
...
+        return createConfig(
+            this.checkRequirements(filesFromDiff),
+            this.composerJsonFileName,
+            this.composerLockJsonFileName,
+            this.continousIntegrationConfigurationJsonFileName
+        );
+    }
+
+    private projectContainsValidJsonConfiguration(schemaName: PathLike, fileName: PathLike): ValidationResult {
+        const schemaPath = `${this.action.getApplicationDirectory()  }/${  schemaName}`;
```
internalsystemerror: Spacing inside the curly braces here seems strange.

src/app.ts
```diff
@@ -0,0 +1,175 @@
...
+        const jobs: [Job, ...Job[]] = (new DefaultJobCreator(tools))
+            .createJobs(config, appConfig);
+
+        this.action.debug(`Jobs created: ${JSON.stringify(jobs, null, SPACES_TO_INDENT_JSON)}`);
+
+        return jobs.map((job) => createJobForMatrixFromJob(job)) as [JobForMatrix, ...JobForMatrix[]];
+    }
+
+    createConfiguration(filesFromDiff: string[]): Config {
+        return createConfig(
```
internalsystemerror: Suggestion, `createConfig` to `createConfigObject`.

src/app.ts
```diff
@@ -0,0 +1,175 @@
...
+
+        return false;
+    }
+
+    private checkRequirements(diff: string[]): Requirements {
+        const CODE_CHECKS_ENABLED_MESSAGE = '- Enabling code checks';
+
+        if (diff.length === 0) {
+            return {
+                codeChecks : true,
```
internalsystemerror, line 116: If there are no changes, run all checks. But if there are changes, run checks selectively based on what has changed. I assume this is deliberate, as my first thought is that these should be false here.

src/app.ts
```diff
@@ -0,0 +1,175 @@
...
+                }
+
+                if (docLintingRegularExpressions.some((expression) => expression.test(filename))) {
+                    requireDocLinting = true;
+                }
+            }
+        );
+
+        if (requireCodeChecks) {
+            this.action.info(CODE_CHECKS_ENABLED_MESSAGE);
```
internalsystemerror: This constant is only used in one place, it could be hard coded. Otherwise, it might be worth moving all messages to a dedicated file (either in an object, or as constants) and including them from there?

src/config/app.ts
```diff
@@ -0,0 +1,328 @@
...
+        phpIni: job.ini ?? appConfig.phpIni,
+        composerDependency: composerDependency,
+        ignorePhpPlatformRequirement: discoverIgnorePhpPlatformRequirementForJobByVersion(job, phpVersion, appConfig),
+        additionalComposerArguments: discoverAdditionalComposerArgumentsForCheck(job, appConfig)
+    };
+}
+
+function convertJobFromFileToJobs(job: JobFromFile, appConfig: Config): Job[] {
+    const jobDefinition: JobDefinitionFromFile = job.job;
+    const composerDependencies = discoverComposerDependenciesForJob(jobDefinition);
```
internalsystemerror: `composerDependencies` confused me for a while. Something like `composerRestraints` or `composerDependencyRestraints` may be more descriptive as with dependencies I think, individual required packages.

src/config/app.ts
```diff
@@ -0,0 +1,328 @@
...
+    });
+
+    return jobs;
+}
+function isJobExcluded(job: Job, exclusions: JobToExcludeFromFile[]) {
+    if (exclusions.length === 0) {
+        return false;
+    }
+
+    return exclusions
```
internalsystemerror: You could use `some` again here instead of `filter` and the length check.

src/config/app.ts
```diff
@@ -0,0 +1,328 @@
...
+    return {
+        name: name,
+        action: ACTION,
+        operatingSystem: OPERATING_SYSTEM,
+        job: {
+            command: command,
+            php: phpVersion,
+            phpExtensions: config.phpExtensions,
+            phpIni: config.phpIni,
+            composerDependency: composerDependency,
```
internalsystemerror: `composerDependency` as with above I think conveys the wrong impression here.

src/config/app.ts
```diff
@@ -0,0 +1,328 @@
...
+            phpIni: [],
+            ignorePhpPlatformRequirement: appConfig.ignorePhpPlatformRequirements[appConfig.stablePhpVersion] ?? false,
+            additionalComposerArguments: appConfig.additionalComposerArguments,
+        }
+    };
+}
+
+export function gatherChecks(config: ConfigurationFromFile, appConfig: Config, tools: Tool[]): [Job, ...Job[]] {
+    if (isExplicitChecksConfiguration(config)) {
+        const checks = (config.checks?.map((job) => convertJobFromFileToJobs(job, appConfig)) ?? []).flat(1);
```
internalsystemerror: Is there a chance for duplicates here? And if so is that desired?

src/config/app.ts
```diff
@@ -0,0 +1,328 @@
...
+            return [createNoOpCheck(appConfig)];
+        }
+
+        return checks as [Job, ...Job[]];
+    }
+
+    let checks = createChecksForKnownTools(appConfig, tools);
+
+    if (isAdditionalChecksConfiguration(config)) {
+        checks = checks.concat((config.additional_checks?.map((job) => convertJobFromFileToJobs(job, appConfig)) ?? []).flat(1));
```
internalsystemerror: As with above regarding duplicate array values.

src/config/input.ts
```diff
@@ -0,0 +1,79 @@
...
+export interface ConfigurationFromFile {
+    extensions?: string[];
+    ini?: string[];
+    ignore_php_platform_requirements?: IgnorePhpPlatformRequirements;
+    stablePHP?: string;
+    additional_composer_arguments?: string[];
+}
+
+export interface JobExclusionsFromFile {
+    exclude?: [JobToExcludeFromFile, ...JobToExcludeFromFile[]];
```
internalsystemerror, line 17: `[JobToExcludeFromFile, ...JobToExcludeFromFile[]]` vs `JobToExcludeFromFile[]`?

src/config/input.ts
```diff
@@ -0,0 +1,79 @@
...
+}
+
+export interface AdditionalChecksConfigurationFromFile extends ConfigurationFromFile, JobExclusionsFromFile {
+    additional_checks?: [JobFromFile, ...JobFromFile[]];
```
internalsystemerror, line 21: `[JobFromFile, ...JobFromFile[]]` vs `JobFromFile[]`?

src/config/input.ts
```diff
@@ -0,0 +1,79 @@
...
+}
+
+export interface ExplicitChecksConfigurationFromFile extends ConfigurationFromFile {
+    checks?: [JobFromFile, ...JobFromFile[]];
```
internalsystemerror, line 25: as above, just wondering why `[JobFromFile, ...JobFromFile[]]` vs `JobFromFile[]`?

src/config/input.ts
```diff
@@ -0,0 +1,79 @@
...
+    job: JobDefinitionFromFile;
+    operatingSystem?: string;
+    action?: string;
+}
+
+export interface JobDefinitionFromFile {
+    php?: InstallablePhpVersionType | AnyPhpVersionType | LowestPhpVersionType | LatestPhpVersionType;
+    extensions?: string[];
+    ini?: string[];
+    dependencies?: AnyComposerDependency | ComposerDependency;
```
internalsystemerror: Suggestion, `depedencyVersionRestraint` or `dependencyRestraint` as noted earlier.

src/config/php.ts
```diff
@@ -0,0 +1,28 @@
...
+    PHP_72,
+    PHP_73,
+    PHP_74,
+    PHP_80,
+    PHP_81
+] as const;
+
+export type InstallablePhpVersionType = typeof INSTALLABLE_VERSIONS[number];
+
+export function isInstallableVersion(version: string): version is InstallablePhpVersionType {
```
internalsystemerror, line 29: is the version provided here always in the `x.x` format, with the patch number excluded?

src/jobs.ts
```diff
@@ -0,0 +1,21 @@
+import { ConfigurationFromFile } from './config/input';
+import { Config, gatherChecks, Job } from './config/app';
+import { Tool } from './tools';
+
+export interface JobCreator {
+    createJobs(config: ConfigurationFromFile, appConfig: Config): [Job, ...Job[]];
```
internalsystemerror, line 7: `[Job, ...Job[]]` vs `Job[]`, I'm sure you have a good reason for this and I'll feel very silly once I find out what it is and will probably start using it going forward.

src/main.ts
```diff
@@ -0,0 +1,50 @@
...
+}
+
+const app = new App(action);
+
+if (!app.sanityChecksPassing()) {
+    process.exit(1);
+}
+
+/* eslint-disable-next-line no-magic-numbers */
+const FIRST_TWO_ARGUMENTS = 2;
```
internalsystemerror: Why make this a constant? As far as I can see this is only used once (and the value is in the name which suggests this wouldn't be changing). It does make it more descriptive as to what's happening, but that could be accomplished with just a comment.

src/tools.ts
```diff
@@ -0,0 +1,131 @@
...
+export type Tool = {
+    executionType: ToolExecutionType,
+    toolType: ToolType,
+    name: string;
+    command: string;
+    filesToCheck: PathLike[]
+}
+
+function detectInfectionCommand(): string {
+    const composerJson: ComposerJson = parseJsonFile('composer.json', true) as ComposerJson;
```
internalsystemerror, line 34: Could this not be reused from elsewhere? I expect hitting the filesystem to be the largest bottleneck in this.

.github/workflows/continuous-integration.yml
boesing: Sure, why not.

Dockerfile
boesing: TBH: I don't know how. If you have concrete suggestions, feel free to drop them here.

composer.schema.json
boesing: Yah, was already thinking about fetching the schema. But this might be outdated since we do not build the matrix when a new composer version is released.
Same for the container.

package.json
boesing: Ah, what do I have to require?
It actually builds so I havent experienced any problem - but I guess thats because something else is requiring typescript for me?

src/config/input.ts
boesing: Should contain at least 1 element.
The other annotation also allows no element.

src/config/input.ts
boesing: non-empty-list

src/config/input.ts
boesing: There is no way that we change input config. I do not want to introduce BC breaks just because we are discussing namings.
Input JSON and output JSON is not arguable. We either have to life with it or start deprecating stuff. But the benefit in this case is not really huge, is it?

src/config/input.ts
boesing: non-empty-list

src/config/app.ts
boesing: since these are explicit checks, I don't care about duplicates.
If the project wants to execute the same check multiple times, thats not our problem?

src/app.ts
boesing: I am no typescript dev. Not sure if we have to make that kind of micro-optimizations either.
But if we want to do it like this, please make a suggestion as I am happy that it works for the moment 😂

src/app.ts
boesing: non-empty-list

src/app.ts
boesing: ❗ 
Good finding, thats from an earlier eslint fix. Have to fix that manually. Will do.

src/app.ts
boesing: selective changes are only made when `diff` is passed.
If there is no `diff` available (due to pushes to a branch for example), we have to execute all checks.

src/app.ts
boesing: ❗ 
Oh yah, it was used more than once before my refactoring. Now its obsolete, true

src/config/app.ts
boesing: Have to google to see what "restraints" is translated to. 😂 

I do not have a meaning here, can change this if we want to rename the initial naming.
I tried to keep as much old stuff as possible since I do not want to totally rework this but first move it to typescript.

src/config/app.ts
boesing: What is `some`? Again: I am no typescript/JS dev.

src/config/app.ts
boesing: Again: if the project somehow manages to declare tests multiple times, who cares? I think de-duplication of checks which are partly generated from wildcard stuff and partly hardcoded would be a bigger pain.
If you know a smart solution, we can optimize this after this patch.

As of today (the non-typescript version), there is no deduplication at all and thus I added some in this PR for things like PHP extensions, etc. but these were all just strings.

src/config/php.ts
boesing: ❗ 
Yes, thats at least how the documentation tells the user to annotate.
We can also enforce this via some kind of JSONSchema `pattern`. 

Marking this as a TODO.

src/app.ts
boesing: ❗

src/jobs.ts
boesing: non-empty-list

src/main.ts
boesing: Because a constant reflects the meaning of it. When reading `arguments.slice(2)`, thats not as verbose as `arguments.slice(FIRST_TWO_ARGUMENTS)`.

Can remove this, don't have a strong feeling for this tho.

src/tools.ts
boesing: I don't think that we have to do these kinds of performance improvements and I would prefer not to pass the composer file around the application.

Anyways, thats how it is done right now - we can optimize this somewhere else but for now I would try to focus on the initial idea: improve dev experience when working on our actions.

If we want to do performance improvements, lets do so in dedicated PRs.

.github/workflows/continuous-integration.yml
internalsystemerror: Adding:
```
      - 'renovate/*'
```
should suffice, or this could be added in a future PR dedicated to renovate.

Dockerfile
internalsystemerror: It means having 2 `FROM` clauses, the first layer being where the application is compiled, and the second layer just being node with the compiled application copied across (node_modules wouldn't be required). (see https://docs.docker.com/develop/develop-images/multistage-build/#use-multi-stage-builds for more).

@Ocramius Pinging you, as you understand the caching better than I do. Is it worth it? The end result of the layers should be the same, as everything within `/usr/local/source` is removed.

composer.schema.json
internalsystemerror: It being outdated wouldn't be a change from now but this might be one for a future refactor as it works as is.

package.json
internalsystemerror: Just add in a dependency to:
```
    "typescript": "~4.6.3"
```
With the `~` constraint instead of `^` since typescript can/does introduce BC breaks in minors.

src/app.ts
internalsystemerror: If you're happy, I'm happy 😄.

src/config/app.ts
internalsystemerror: restraint or constraint, just to imply that this is adding a restriction to the dependency resolution. I wasn't aware the name dependencies was part of the configuration which we don't want to change though, so I'll give way to others opinions here, but when reading through the code, I was expecting this to be a list of required packages.

src/config/app.ts
internalsystemerror:
```ts
return exclusions
        .filter((exclude) => job.name === exclude.name)
        .length > 0;
```

could be written as:
```ts
return exclusions
        .some((exclude) => job.name === exclude.name);
```

it just returns a boolean if any of the values in the array return true for the callable.

src/config/app.ts
internalsystemerror: Fair

src/config/app.ts
internalsystemerror: Fair. Happy to give way here. If you do wish to add deduplication, an easy way is to:
```ts
test.filter((value, index, array) => array.indexOf(value) === index);
```

src/config/input.ts
internalsystemerror: Ah yes, of course.

src/config/input.ts
internalsystemerror: Yep, my mistake.

src/config/input.ts
internalsystemerror: Completely fair, I wasn't thinking about this being used outside of this project.

src/main.ts
internalsystemerror: I understand, I just would have used a comment if you felt it was confusing.

```
const FIRST_TWO_ARGUMENTS = 2;
const config = app.createConfiguration(process.argv.slice(FIRST_TWO_ARGUMENTS));
```

vs

```
// Create configuration using the first 2 CLI arguments
const config = app.createConfiguration(process.argv.slice(2));
```

These are functionally equivalent though, so happy to go with whatever you prefer.

src/tools.ts
internalsystemerror: Fair.

tsconfig.json
```diff
@@ -0,0 +1,20 @@
+{
+    "compilerOptions": {
+        "module": "commonjs",
+        "target": "esnext",
+        "lib": ["es2015", "es2017"],
+        "strict": true,
+        "alwaysStrict": true,
+        "esModuleInterop": true,
+        "skipLibCheck": true,
```
internalsystemerror, line 10: Could you also add here:
```
    "noUncheckedIndexedAccess": true,
```
as this isn't included in the `strictNullChecks` which is enabled via `"strict": true"`

boesing: Sure, why not. Will add it with the other fixes.
❗

src/app.ts
boesing: Well, I am happy with simple things. Not yet a real pro with typescript and thus not too familiar with best practices 😅
Havent spent any time in stuff like this - if you dont mind, I would love to understand the idea behind this so I can improve it and also keep that in mind for future projects 👌🏻

src/config/app.ts
boesing: Well, as a `constraint` I would assume we are talking about these `~8.0.0` or `^7.4 || ^8.0`, but actually we are talking about `locked`, `latest` and `lowest` dependencies.

.github/workflows/continuous-integration.yml
boesing: Yah, I'd love to have a dedicated renovate PR so we can focus on typescript here for now.

src/action/github.ts
boesing: I've removed the interface extension and instead added a method `Action#getLogger`. This will actually return the logger so that can be injected and used instead.

src/config/app.ts
boesing: I've renamed it to `dependencySet` where possible.
This might reflect the meaning a little bit better as this will tell composer which dependency set to install:
- `locked` dependencies from `composer.lock`
- `latest` dependencies which are supported by the projects constraints
- `lowest` dependencies which are supported by the projects constraints

src/app.ts
internalsystemerror: ~~My assumption is based on the way javascript executes in that it "hoists" variables during the initial creation context where they are first set to be undefined, and then the value calculated whereas consts are defined in a single execution. JS changes often though so I'm not sure how relevant that still is.~~
After a bit of researching, I think this only applies to `var` not `let`.

Dockerfile
boesing: I reworked the build. Not sure if we already make use of any cache here (guess not) but could be done in a dedicated ticket anyway

composer.schema.json
boesing: Changed the dockerfile to actually receiving the composer schema from `getcomposer.org` while building the container

internalsystemerror (approved):
I am appeased. This is some good work @boesing.

.github/workflows/continuous-integration.yml
```diff
@@ -0,0 +1,38 @@
...
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v1
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+
+      - name: Build
+        uses: docker/build-push-action@v2
+        with:
+          push: false
```
Ocramius, line 120: Nice! One note: we have no caching on this, so that will probably be a bit slow, but this component changes rarely (urgh, too rarely, sorry!)

src/action/local.ts
```diff
@@ -0,0 +1,35 @@
+import {PathLike} from 'fs';
+import {Action} from '../action';
+import {Output} from '../config/output';
+import {SPACES_TO_INDENT_JSON} from '../json';
+import { Logger } from '../logging';
+
+export class Local implements Action {
```
Ocramius, line 6: Don't expect users to run this locally, but is not important.

src/app.ts
```diff
@@ -0,0 +1,175 @@
...
+        return createConfigObject(
+            this.checkRequirements(filesFromDiff),
+            this.composerJsonFileName,
+            this.composerLockJsonFileName,
+            this.continousIntegrationConfigurationJsonFileName
+        );
+    }
+
+    private projectContainsValidJsonConfiguration(schemaName: PathLike, fileName: PathLike): ValidationResult {
+        const schemaPath = `${this.action.getApplicationDirectory()}/${schemaName}`;
```
Ocramius: Very convoluted way to concatenate two strings, but okay O_o

src/config/app.ts
```diff
@@ -0,0 +1,353 @@
...
+    }
+
+    const composerPhpVersion: string = (composerJson.require?.php ?? '').replace(/,\s/, ' ');
+
+    if (composerPhpVersion === '') {
+        return [];
+    }
+
+    return INSTALLABLE_VERSIONS
+        .filter((version) => semver.satisfies(`${version}.0`, composerPhpVersion));
```
Ocramius, line 31: Hmm, probably inherent from original design, but this means we need to touch this action at each release 🤔 

Non-issue for porting to TS, just a problem for the future.

src/config/app.ts
```diff
@@ -0,0 +1,353 @@
...
+        latestPhpVersion              : maximumPHPVersion,
+        phpIni                        : configurationFromFile.ini ?? [],
+        lockedDependenciesExists      : fs.existsSync(composerLockJsonFileName),
+        exclude                       : gatherJobExclusions(configurationFromFile),
+        ignorePhpPlatformRequirements : configurationFromFile.ignore_php_platform_requirements ?? {},
+        additionalComposerArguments   : [ ... new Set(configurationFromFile.additional_composer_arguments ?? []) ],
+    };
+}
+
+export type IgnorePhpPlatformRequirements = Partial<Record<typeof INSTALLABLE_VERSIONS[number], boolean>>
```
Ocramius, line 464: Really cool way to declare a partial :O

src/config/composer.ts
```diff
@@ -0,0 +1,20 @@
+export interface ComposerRequirements {
+    [key: string]: string
+}
+
+/**
+ * @title Package
```
Ocramius: Not needed?

src/json.ts
```diff
@@ -0,0 +1,33 @@
+import fs, { PathLike } from 'fs';
+
+export const SPACES_TO_INDENT_JSON = 2;
+
+export function parseJsonFile(jsonFile: PathLike, allowMissing: boolean = true): object {
```
Ocramius, line 5: eslint doesn't seem to like this: weird.

Still, LGTM.

src/json.ts
```diff
@@ -0,0 +1,33 @@
...
+        parsedJsonFile = JSON.parse(fs.readFileSync(jsonFile).toString());
+    } catch (error) {
+        let failedReason = JSON.stringify(error);
+
+        if (error instanceof SyntaxError) {
+            failedReason = `${error.message}`;
+        }
+
+        throw new Error(`Failed to parse ${jsonFile}: ${failedReason}`);
+    }
```
Ocramius, line 22: Wouldn't it suffice for the json exception to bubble up?

Ocramius (approved):
FML, this has been open for over 6 months? Really sorry for that, @boesing :-( 

I took a slow read through it, my TS-fu is low, but I didn't see any particular scary usages of casts and `any` or `unknown`, so IMO 🚢  at your will.

I will gladly try out the new version on my repos!

src/app.ts
internalsystemerror: This is pretty typical in TS

froschdesign: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals

src/action/local.ts
boesing: This is mainly for development. Without this, its almost impossible to debug stuff. I spent weeks working on this component, I know what I am talking about ;-)

src/app.ts
boesing: Its also enforced by `tslint`

.github/workflows/continuous-integration.yml
boesing: Yah I know but would be fine for me. This is just for CI so we can actually use the final image to run integration tests on and afaik its done only once. 
The dependencies are also very limited and thus its lightning fast.

src/config/composer.ts
boesing: It was when I created the `composer.schema.json` from this file. But I dropped that idea but did not deleted this annotation. So yes, I will delete it. Thanks!

src/config/app.ts
boesing: Yes, on each PHP release.
Once.

I don't see any other possibility for projects using `>= 5.6`. How shall we know which PHP versions we do support?
This is also a task because we also have to manually touch the container action to compile that PHP version into the container. So this would be a direct follow-up anywas.

src/json.ts
boesing: This file here is re-used and thus, used for multiple JSON files. Thats why I wanted to add the JSON file path to the error here.

src/json.ts
internalsystemerror: It's because the boolean ia already inferred from the = true, so unless the type could be `boolean|object` for instance, it's saying you may aswell drop it.

Ocramius: If this is normal in typescript (default parameter value restricting parameter type), then no problem :+1:

boesing: I disabled this in the `eslintrc` because heck, I do not want them to "infer" it here. Being more strict is always better IMHO and thus I disabled this at all.

src/app.ts
```diff
@@ -37,7 +37,7 @@ export class App {
         );
 
         if (!schemaValidationResult.valid) {
-            return this.failWithLoggedJsonSchemaValidationErrors(
+            this.failWithLoggedJsonSchemaValidationErrors(
```
internalsystemerror, line 40: Doesn't this change mean that `sanityChecksPassing()` always returns true?

boesing: Good catch.

I've changed the return type to `void` and handled that in `main.ts`.
Since the `never` methods already handle `exit` on themselves, the `exit` handling within the `main.ts` is not necessary anymore.

Ocramius:
This is starting to look more green-ish 💪

src/app.ts
boesing: Fixed here: https://github.com/laminas/laminas-ci-matrix-action/pull/83/commits/59ee75d61e8c337fed510701f81175406cae3b58

Ocramius (approved):
IMO 🚢

PR #117, opened 2022-08-06
Update required node version to match the one installed

internalsystemerror:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | yes

### Description

Dockerfile installed node v18, however package.json is using node v17. This PR updates the latter to ensure there is no confusion by developers or by the tools.

There are also a couple of removal instances of unused lines of code, which I can undo if requested. In this vein, I also added double quoting inside the shell scripts to prevent globbing and word splitting.

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -54,14 +54,14 @@ function checkout {
 
     if [[ "$REF" == "$LOCAL_BRANCH" ]];then
         echo "Checking out ref ${REF}"
-        git checkout $REF
+        git checkout "$REF"
     else
         echo "Checking out branch ${BASE_BRANCH}"
-        git checkout ${BASE_BRANCH}
+        git checkout "${BASE_BRANCH}"
         echo "Fetching target ref ${REF}"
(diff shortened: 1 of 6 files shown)


internalsystemerror:
The target/lib version of ECMAscript to use was taken from: https://github.com/microsoft/TypeScript/wiki/Node-Target-Mapping. Node v18 will _probably_ support `es2022`, however https://node.green/#ES2022 reports that this is only 97% implemented.

Ocramius (approved):
🚢 build is also green

Ocramius:
Well, post-merge build is red 😩

Ocramius:
Nvm, seems like a `docker pull` limit issue, unrelated to this

internalsystemerror:
Odd that it was green in the PR.

I suspect you're correct, but https://github.com/laminas/laminas-ci-matrix-action/tree/1.14.x/tests/code-check-deprecated-exclusion-via-config is the failing test on both release branches.

PR #119, opened 2022-08-10
Feature: Use `composer/semver` to detect PHP version ranges

### LINKED ISSUE #86: Composer `php` constraints with `,` are not supported
### Bug Report

|    Q        |   A
|------------ | ------
| Version(s)  | 1.11.5

#### Summary

When a component uses a more complex constraint which contains `,`, the matrix is not able to infer supported PHP versions anymore.

#### Current behavior

Supported PHP versions are not detected at all.

#### How to reproduce

*composer.json*
```json
{
    "require": {
        "php": ">=5.6,<=8.1.99"
    }
}
```

#### Expected behavior

All versions between PHP 5.6 and 8.1.99 are detected. As of writing this issue, this would be:
- 5.6
- 7.0
- 7.1
- 7.2
- 7.3
- 7.4
- 8.0
- 8.1

### LINKED ISSUE #87: Use `composer/semver` internally to verify PHP version compatibility
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | yes
| BC Break    | no

#### Summary

As of now, we are using some semver javascript library. Since that actually does not align with all the "features" of `composer`, we might want to create a little PHAR which uses `composer/semver` under the hood to simply exit with `0` if PHP version is supported by the constraint or `1` if not.

Adding some kind of `child_process` to our matrix javascript node binary and we can run that PHAR from within JS for a more accurate version detection.

This would also fix #86 

Thoughts?

boesing:
|    Q          |   A
|-------------- | ------
| Bugfix        | yes
| BC Break      | no
| New Feature   | yes

### Description

By using composers own `composer/semver` library, we are able to detect PHP version ranges from almost every composer-compatible constraint.

Fixes #87


Dockerfile
```diff
@@ -25,6 +27,16 @@ ADD laminas-ci.schema.json /action/
 COPY --from=compiler /usr/local/source/dist/main.js /action/
 RUN chmod u+x /action/main.js
 
+# Setup PHP
```
boesing, line 30: I know, what the fuck. But since I am not aware on how to actually merge multiple container together, I've thought I can do it like this.

This will:
1. ensure that renovate is able to bump versions
2. provide alpine PHP version within this container
3. allows us to run PHP natively within the matrix container while the container itself is optimized for `nodejs`

I hope we can keep it like this as I tried to install `php8-cli` via `apk` which is PHP 8.1.1 which is not supported by the "boxed" PHAR of my `composer-semver` docker image...

package.json
```diff
@@ -25,7 +25,6 @@
   "dependencies": {
     "@actions/core": "^1.6.0",
     "@cfworker/json-schema": "^1.12.2",
-    "@types/semver": "^7.3.9",
-    "semver": "^7.3.5"
+    "child_process": "^1.0.2"
```
boesing, line 28: I have no idea where this comes from, should I drop it?

src/config/app.ts
```diff
@@ -16,19 +16,33 @@ export enum ComposerDependencySet {
     LATEST = 'latest',
 }
 
-function gatherVersions(composerJson: ComposerJson): InstallablePhpVersionType[] {
+function gatherVersions(composerJson: ComposerJson, logger: Logger): InstallablePhpVersionType[] {
```
boesing, line 19: I do still not really know how to refactor this to have actual DI.
Maybe @internalsystemerror has some ideas on how to refactor this to make it more OOP.

package.json
boesing: NVM, it seems that I added it due to the `execSync`.

package.json
```diff
@@ -5,9 +5,9 @@
   },
   "type": "commonjs",
   "devDependencies": {
+    "@types/node": "^18.6.4",
```
internalsystemerror, line 8: did you move this or a tool? Only I've seen tools move it the other way around before.

Dockerfile
internalsystemerror: ~~Does the matrix container run any PHP? What happens with PHP 8.0 or PHP 7.4 runs?~~ nm, I see where it's used now

src/config/app.ts
internalsystemerror: I'm mostly familiar with the DI in Angular, but there are (smaller) libraries available e.g. [TypeDI](https://github.com/typestack/typedi) or even [TSyringe](https://github.com/microsoft/tsyringe) by MS

src/semver.ts
```diff
@@ -0,0 +1,11 @@
+import {execSync} from 'child_process';
+
+export function satisfies(version: string, constraint: string): boolean {
+    try {
+        execSync(`/usr/local/bin/composer-semver.phar semver:match "${version}" "${constraint}" > /dev/null`);
```
internalsystemerror, line 5: There are some edge case differences between how composer and npm resolve semantic versions (or so I'm led to believe), but I don't see why we aren't using the official semver module used by npm https://www.npmjs.com/package/semver and not even require PHP here

internalsystemerror: To clarify, I know this PR is removing it and replacing it with composer, I'm just wondering why?

internalsystemerror:
Everything checks out though. Looks good to merge if we decide to go down this route, but I'm not convinced its necessary.

package.json
boesing: I guess this was due to a merge conflict. So maybe I did that.

src/config/app.ts
boesing: Noted. Created an issue based on your comment. Thanks!

src/semver.ts
boesing: Since we do parse `composer.json`, that file could provide constraints which are mostly supported by composer but i.e. not by the NPM package.

I have not tested this yet, but composer provides some quirks:
- `^v8.1.0` would be a valid constraint
- Comma annotations are valid constraints (as outlined in #86)

There might be other quirks but since I do not want to hassle with these in this component since we should focus on detecting the correct supported version(s) based on what is supported by composer, we should use composers own logic to detect whats allowed and what not.
But thats just my opinion here.

internalsystemerror: That first one would be a valid constraint in node aswell FYI. I did look through the docs for both npm and composer and the comma vs space was the only one I found. Which IMO is npm's mistake here as I know of other examples where a comma is perfectly valid (e.g. go).

boesing: I have no strong opinion on this, I just would prefer not to manipulate the constraint to "make it work" 😅 
But I am open for every other suggestion. Should we report this upstream?

internalsystemerror: Valid. I also see the case for using composer directly. It's a good argument. I just have an uncomfortable feeling about mixing services within a single container, and more than that mixing languages.

Maybe we should consider porting this to PHP instead 🤣. (some day maybe)

boesing: TBH: I was surprised that @weierophinney wrote this in javascript in the first place. I guess that was due to the fact that github provides [a library](https://www.npmjs.com/package/@actions/core) to actually interact with the GHA "API".

weierophinney: I wrote it in JS for a few reasons:

- I'm reasonably proficient in it already.
- The base workflow "container" that GHA provides has JS available out-of-the-box, and includes the GHA API (as @boesing notes).
- I'd not had a chance (and still have not!) to learn TypeScript, and didn't feel I had the luxury of doing that in order to put this together (was trying to address a very real pain point we had with CI, and had limits both in the time I had available, and the amount of time Marco had then to review and help me revise).

I'm happy to see this move to TS; I think it gives us a better safety net. Unfortunately, it also means that for now, I'm not really able to review and provide meaningful feedback!

weierophinney: Regarding the "why choose JS", part of the decision was _performance_. One of the requirements we put for ourselves was that generation of the matrix should be **fast** - as in, no more than 3 seconds. The only way we could get that sort of speed with PHP would be if we created a container for the action with minimal dependencies... and I wasn't sure we could do that reasonably. Today, I think we could.

internalsystemerror: @weierophinney do you have any thoughts at an abstract level on an alternative approach to call `composer/semver` from the matrix app? The method implemented here is to include PHP in a node container and call semver via `execShell`.

Dockerfile
```diff
@@ -25,6 +27,16 @@ ADD laminas-ci.schema.json /action/
...
+RUN mkdir -p /usr/local/bin /usr/local/etc /usr/local/lib
+COPY --from=php /usr/local/bin/php /usr/local/bin
+COPY --from=php /usr/local/etc/* /usr/local/etc
+COPY --from=php /usr/local/lib/php/* /usr/local/lib/php
+COPY --from=php /usr/lib/* /usr/lib
+COPY --from=php /lib/* /lib
```
internalsystemerror, line 36: it's these parts that worry me mixing the languages. if they both rely on a linked lib in different versions for instance.

src/semver.ts
internalsystemerror: what about calling the php container via docker here instead? I'm not that familiar with how DinD works on github specifically but something like:

`execShell('docker run php /usr/local/bin/composer-semver.phar semver:match "${version}" "${constraint}" > /dev/null')`. After adding `composer-semver` to the php container instead?

weierophinney: Won't work; you're operating _inside_ a container, so the container can not operate as a docker host.

As far as other suggestions... I don't have any. I used the NPM semver package originally, as I assumed it was compatible with how Composer handled semver. Knowing that they're different, we can either:

- Call on a PHP executable (as proposed here)
- Write our own TS utility that passes the same tests as composer/semver (lots more effort, plus it can go out-of-date)

It might make sense to re-do this action such that it uses a dedicated container where we use a PHP application, as we've done with the CI action and the automatic releases action.

boesing: > Won't work; you're operating inside a container, so the container can not operate as a docker host.

Thats actually not true. It definitely works, I am just not sure if it works within GHA.
One has to mount the docker socket to the container tho.

boesing: I am still wondering if we should just report upstream that they can't handle `,` in the constraint?
Maybe its a fixable issue for them?

Dockerfile
Ocramius: Similar thoughts: would prefer a `composer install` happening with our base image, but we can improve on that later

tests/issue-86/.laminas-ci.json
```diff
@@ -0,0 +1,11 @@
+{
```
Ocramius, line 1: instead of `issue-86`, would it make sense to name this according to what it does?

src/semver.ts
internalsystemerror: Issue opened https://github.com/npm/node-semver/issues/468, we can see what they say

Dockerfile
boesing: There is no need for `composer install`, thats actually to execute the `composer-semver.phar` I am building with https://github.com/boesing/composer-semver-docker

But the more I think about this, the more it annoys me that it has to be that hacky...

boesing:
I've raised an "issue" in the NPM semver implementation:
https://github.com/DefinitelyTyped/DefinitelyTyped/discussions/61669

internalsystemerror:
As I understand it, that repository is for the typescript mappings for everything `@types/` related, including `@types/semver`. The implementation for the `semver` library would be https://github.com/npm/node-semver. I've opened an issue for that one here https://github.com/npm/node-semver/issues/468. 🤞

internalsystemerror:
It appears there are [some other differences](https://github.com/semver/semver/pull/584#issuecomment-705503374), so I think we have to go this route after all.

PR #122, opened 2022-08-19
Revert "Use `XmlLint` against tool configurations (PHPUnit, PHPCS, Psalm)"

Ocramius:
Reverts laminas/laminas-ci-matrix-action#31

As discussed with @internalsystemerror on slack, this didn't work as expected, since it leads to failing builds all over the place.

Reverting for now, and will be re-applied to `1.15.x` in future, once ready.

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,6 @@
   },
   "scripts": {
     "lint": "eslint src/ --ext .ts",
-    "lint-fix": "eslint src/ --ext .ts --fix",
     "build": "tsc --build && webpack --mode=production"
   },
   "dependencies": {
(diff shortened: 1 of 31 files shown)


internalsystemerror:
This revert can be reverted once https://github.com/laminas/laminas-continuous-integration-action/pull/123 is merged.

@Ocramius is there a way to ensure we require the correct version of that in this when it is?

Ocramius:
I think it always uses the latest version, which is a bit of a pain...

boesing:
Monorepo will come - when I find some more time and passion to work on this.

internalsystemerror:
Fix has been released as https://github.com/laminas/laminas-continuous-integration-action/releases/tag/1.24.0. This revert can be reverted.

PR #126, opened 2022-09-03
WIP: Verify BC breakages on pull requests, by comparing PR against base branch

Ocramius:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | yes
| RFC           | no
| QA            | yes


.github/workflows/continuous-integration.yml
```diff
@@ -84,7 +84,15 @@ jobs:
...
         env:
           PROJECT_NAME_TO_TEST: ${{ matrix.projectName }}
-        run: cd tests/${PROJECT_NAME_TO_TEST} && docker run -i --entrypoint "/action/main.js" -v $(realpath .):/github/workspace -w=/github/workspace ${TEST_TAG} $(test -r diff && cat diff || echo -n "")
+        run: |
+          cd tests/${PROJECT_NAME_TO_TEST} && \
+          docker run \
+            -i \
+            --entrypoint "/action/main.js" \
+            -v $(realpath .):/github/workspace \
+            --env-file=test.env \
```
Ocramius, line 93: I added a `test.env` to each test directory: ugly, but IMO better than adding more conditional logic here

package.json
```diff
@@ -25,7 +25,9 @@
   },
   "dependencies": {
     "@actions/core": "^1.6.0",
+    "@actions/github": "^5.0.3",
     "@cfworker/json-schema": "^1.12.2",
+    "@octokit/webhooks-definitions": "^3.67.3",
```
Ocramius, line 30: This is needed to cast the `context.payload` to a typed data structure

package.json
```diff
@@ -25,7 +25,9 @@
   },
   "dependencies": {
     "@actions/core": "^1.6.0",
+    "@actions/github": "^5.0.3",
```
Ocramius, line 28: This is needed to read `github.context.payload`, which contains the base ref of a PR

src/action/github.ts
```diff
@@ -32,4 +34,14 @@ export class Github implements Action {
             }
         };
     }
+
+    getBaseBranchSha1(): string | null {
```
Ocramius, line 38: I returned `string | null` from most endpoints, but perhaps a `Maybe` type would work better

src/tools.ts
```diff
@@ -41,8 +42,26 @@ function detectInfectionCommand(): string {
...
-    return [
+function backwardCompatibilityCheckTool(action: Action): Tool | null {
+    const baseSha1 = action.getBaseBranchSha1();
+
+    if (baseSha1 === null) {
+        return null;
+    }
+
+    return {
+        // @TODO need to `git fetch baseSha1` from source repo!
```
Ocramius, line 53: Yep, we need a `git fetch` here! 

Unsure how to handle it best, but @basz ran into some `git clone` problems with private repos, so perhaps we can fix that together with it :)

src/tools.ts
```diff
@@ -115,8 +134,15 @@ export default function createTools(config: Config): Array<Tool> {
             command       : './vendor/bin/codecept run',
             filesToCheck  : [ 'codeception.yml.dist', 'codeception.yml' ],
             toolType      : ToolType.CODE_CHECK,
-        }
-    ]
+        },
+    ] as Tool[];
+
+    if (bcTool !== null) {
```
Ocramius, line 147: yuck yuck yuck - would prefer to merge two sets of `Tool[]`, but for now, this will have to do.

tests/code-check-roave-backward-compatibility-check-on-pr/GITHUB_EVENT_CONTENTS.json
```diff
@@ -0,0 +1,328 @@
+{
```
Ocramius, line 1: A github event is **HUGE** :O

src/tools.ts
```diff
@@ -41,8 +42,26 @@ function detectInfectionCommand(): string {
...
+        return null;
+    }
+
+    return {
+        // @TODO need to `git fetch baseSha1` from source repo!
+        executionType : ToolExecutionType.STATIC,
+        name          : 'Backward Compatibility Check',
+        command       : `roave-backward-compatibility-check check --from="${ baseSha1 }" --install-development-dependencies`,
+        filesToCheck  : [ 'composer.json' ],
+        toolType      : ToolType.CODE_CHECK,
```
Ocramius, line 58: Defining `before_script` here would probably be needed.

tests/code-check-roave-backward-compatibility-check-on-pr/test.env
```diff
@@ -0,0 +1,2 @@
+GITHUB_EVENT_PATH=/github/workspace/GITHUB_EVENT_CONTENTS.json
+GITHUB_EVENT_NAME=pull_request
```
Xerkus, line 2: For context, used environment variables come from 

https://github.com/actions/toolkit/blob/bc4be505973a6a7344bfd71e1b32f77e1755310c/packages/github/src/context.ts#L28-L53

@Ocramius should other variables be set to some non-empty dummy value?

Ocramius: At this stage, doesn't look like it.

package.json
```diff
@@ -5,9 +5,9 @@
   },
   "type": "commonjs",
   "devDependencies": {
+    "@types/node": "^18.6.4",
```
internalsystemerror, line 8: Just out of interest, did you move this? I saw a similar change previously. If npm did this then consider this resolved.

src/action/github.ts
internalsystemerror: If for whatever reason the context is not set correctly then it could return undefined. It depends if an error is thrown during casting before it gets to that point.

Alternatively you could avoid null and just return undefined (again depends on how the casting library works if null would ever be allowed). Maybe<string> works for me too if you'd rather that approach.

src/tools.ts
internalsystemerror: Have you tried adding `git fetch ${baseSha1}` as a `before_script`?

src/tools.ts
internalsystemerror: You could do something like:

```ts
const tools = [
    ...[bcTool].filter()
    <other tools>
];
```

Spread new array containing the `bcTool with filter removing null values.

package.json
Ocramius: Yep, used `npm install --save` for the added dependencies

src/action/github.ts
Ocramius: Should be a concern of the upstream library, no?

src/tools.ts
Ocramius: Tried that, but `.filter()` TSC wasn't able to infer `Tool[]` from it, due to too many dynamic expressions, I guess.

src/tools.ts
Ocramius: How do we add a `before_script` though?

src/tools.ts
internalsystemerror: `...([bcTool].filter() as Tool[])` will do that.

src/tools.ts
internalsystemerror: ah yes, at this point I'd added `lintConfigCommand` because the file that exists in `filesToCheck` is appended to it before it gets added to `before_script`.

You could add another option key `beforeScript`, and then modify https://github.com/laminas/laminas-ci-matrix-action/blob/3e17f6852334fef66e6b973a0b263cbacaa94458/src/config/app.ts#L290 to use it?

src/action/github.ts
internalsystemerror: It should be. I've double checked the types and it looks good.

src/app.ts
```diff
@@ -65,7 +65,7 @@ export class App {
         appConfig: Config
     ): [JobForMatrix, ...JobForMatrix[]] {
         const config = parseJsonFile(this.continousIntegrationConfigurationJsonFileName, true) as ConfigurationFromFile;
-        const tools = createTools(appConfig);
+        const tools = createTools(appConfig, this.action);
```
boesing, line 68: Can we pass the hash in the `appConfig` instead?
It also holds if there is a lockfile so we do not have to check for `fileExists(lockFile)` in every tool.
I would prefer having a dedicated object/property in the app config for the target branch REF as well.

src/tools.ts
boesing: Imho, we should move `tool.lintConfigCommand` into the tool itself and provide `beforeScript` from within the tool itself.

tests/code-check-roave-backward-compatibility-check-on-pr/GITHUB_EVENT_CONTENTS.json
boesing: Do we really need the whole event here? is there a check which verifies that the JSON is 100% available?

I would expect that we just pass those parts to the matrix which are actually used. Makes things easier as this event might grow in the future anyways and thus we will never have 100% available.

boesing (changes requested):
Change looks fine for me.

- Passing `git fetch <basebranch>` to `before_script` should work.
  I'd say lets add it from the tool. I would prefer keeping these `before_script` and `after_script` things in internal scope for now. No need to open that up for every check.

- Can we probably remove the empty `test.env` files? I do not see a good reason why we should have a bunch of dead files around when they do not make a difference.

- I would prefer to pass the `targetRef` via the `appConfig` where the supported PHP versions of the project are stored, where the information if a lock file is available is stored, etc. This would probably reduce the LoC which have to change to a minimum and we do not need to pass the action through half of the application scope.

tests/code-check-roave-backward-compatibility-check-on-pr/GITHUB_EVENT_CONTENTS.json
Ocramius: This is the correct event according to its JSON Schema: I wouldn't cut a corner here. 

 > I would expect that we just pass those parts to the matrix which are actually used. 

The structure is based on the octokit type (and associated type cast).

 > Makes things easier as this event might grow in the future anyways and thus we will never have 100% available.

Would rather say that we get a BC break warning heads-up, should they expand it further :D 

So far, new fields are marked optional: I also skipped all optional fields from here.

src/tools.ts
internalsystemerror: Be careful as I did that originally but needed a way to append the filename to the lint command.

src/tools.ts
boesing: maybe having multiple before scripts might help?

1. `git remote add upstream <upstreamUri>`
2. `git fetch upstream <baseRef>`

boesing:
Don't we get the target branch ref from environment?
https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables

- `GITHUB_BASE_REF`

Do we really need that whole event parsing here or should we just grab the ref from environment instead?

`const GITHUB_BASE_REF = process.env["GITHUB_BASE_REF"] || ""` should already do the job, not?

Ocramius:
I can't seem to be able to allocate a chunk of time to this.

It would save me a massive amount of time in code reviews, but debugging the last details (around fetching multiple refs) is not something I'll be able to do soon.

boesing:
I'll see if I find some time and apply a few things.

I would prefer to avoid the whole event stuff and instead use `GITHUB_BASE_REF`. This would tighten down the amount of changes here to a minimum and then we can start with an initial PoC.

We could also release a version where we can opt-in to the new job before rolling out for everyone by-default.

i.e. a `.laminas-ci.json` flag (which can also be used to fully disable the check in case some1 wants to). This will provide us the possibility to have it field tested in some projects for quite a while. Especially the whole "multiple ref" stuff (which I probably did not fully understand right now) could be verified.

In one of the following releases we can then change the default of the `.laminas-ci.json` flag to `true` and rollout the feature for everyone by default.

WDYT?

src/tools.ts
boesing: Might be helpful once this PR is being continued:

https://github.com/symfony/symfony/blob/3dfec48b454e24ad2ee3aace4fb2f05c2109d3c5/.github/workflows/psalm.yml#L32-L35

boesing: I'll leave this here: 
https://github.com/laminas/laminas-continuous-integration-action/blob/fb9164e3cf7fbbd9abacd4d7f7a4a30f165bb7dd/entrypoint.sh#L78-L79

Seems that the CI Container is already checking out the target branch and thus, should be available

src/tools.ts
boesing: https://github.com/laminas/laminas-ci-matrix-action/pull/126#discussion_r1256601809

boesing:
@Ocramius I am unable to push to your fork and thus, can't push my changes.
Could you enable maintainer edits or does the fork has limited access?

boesing:
Superseded by #226

Ocramius:
> Could you enable maintainer edits or does the fork has limited access?

I generally don't do this due to repo secrets :-) 

Anyway, thanks for taking over on this!

PR #136, opened 2022-10-29
Add jest and a simple test example

internalsystemerror:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | yes
| RFC           | no
| QA            | yes

### Description

In order to investigate https://github.com/laminas/automatic-releases/pull/212, and as previously requested by @boesing we need to add unit tests.

In adding a simple example for `config/app.ts#createConfig()` it appears that the `stableVersion` for `8.1` is actually `7.4` which may be the cause of the issue above.


internalsystemerror:
@Ocramius I think we have possibly found the cause of the issue on `automatic-releases` (see https://github.com/laminas/laminas-ci-matrix-action/actions/runs/3352246718/jobs/5554266299).

@boesing I know you've done a lot of work on this, so I wanted to set up unit testing and ping you so you're kept in the loop.

Dockerfile
```diff
@@ -4,7 +4,7 @@ RUN mkdir -p /usr/local/source
 WORKDIR /usr/local/source
 COPY package*.json ./
 COPY tsconfig.json ./
-COPY webpack.config.js ./
+COPY webpack.config.ts ./
```
Ocramius, line 7: Ooooh, didn't know this worked!

package.json
```diff
@@ -1,32 +1,38 @@
...
-    "build": "tsc --build && webpack --mode=production"
-  },
-  "dependencies": {
-    "@actions/core": "^1.6.0",
-    "@cfworker/json-schema": "^1.12.2",
-    "@types/semver": "^7.3.9",
-    "semver": "^7.3.5"
-  }
+    "name": "laminas-ci-matrix-action",
+    "description": "Create a CI matrix for use in a GitHub Action",
```
Ocramius, line 3: Btw, please don't reformat everything next time 😁

package.json
```diff
@@ -1,32 +1,38 @@
...
+        "eslint-config-incredible": "^2.4.2",
+        "eslint-import-resolver-typescript": "^3.5.2",
+        "eslint-plugin-import": "^2.26.0",
+        "eslint-plugin-node": "^11.1.0",
+        "jest": "^29.2.2",
+        "source-map-loader": "^4.0.1",
+        "ts-jest": "^29.0.3",
+        "ts-node": "^10.9.1",
+        "typescript": "~4.8.4",
+        "webpack": "^5.74.0",
```
Ocramius, line 35: Just realized: we don't really webpack anything, do we? 🤔

src/config/app.spec.ts
```diff
@@ -0,0 +1,32 @@
+import {PathLike} from 'fs';
+import createConfig from './app';
+
+describe('config/app', () => {
+    describe('createConfig()', () => {
+        const phpIniFromConfigurationPath: PathLike = 'tests/php-ini-from-configuration';
+
+        it('should return valid config', () => {
+            expect(createConfig(
```
Ocramius, line 29: Overall okay with this testing approach, but note that we could've caught this with the previous testing approach too 👍

src/config/app.spec.ts
```diff
@@ -0,0 +1,32 @@
...
+                {
+                    codeChecks : true,
+                    docLinting : true,
+                },
+                `${phpIniFromConfigurationPath}/composer.json`,
```
Ocramius, line 34: What are we testing here? 🤔

Would probably make sense to have these files readable inside the test?

src/config/app.spec.ts
```diff
@@ -0,0 +1,32 @@
...
+                `${phpIniFromConfigurationPath}/composer.lock`,
+                `${phpIniFromConfigurationPath}/.laminas-ci.json`
+            )).toEqual({
+                codeChecks                    : true,
+                docLinting                    : true,
+                versions                      : [ '8.1' ],
```
Ocramius, line 40: One thing I realized: `~8.1.2` will, for example, never match `8.1` 🤔

package.json
internalsystemerror: Yeah I wouldn't normally, but the vast majority of repos have `scripts` at the top, and I may have gone a bit too far :(

package.json
internalsystemerror: The `build` command in scripts runs `tsc --build && webpack --mode=production`, which is called from `Dockerfile`.

src/config/app.spec.ts
internalsystemerror: I think of the current approach as higher up the pyramid. Unit testing will give us quicker clarity as to what the issue is, so the more the better I say.

src/config/app.spec.ts
internalsystemerror: Yeah, this was more just a quick example, but it makes sense to have distinct tests. Ideally I'd have passed in the configurations themselves, but these needed filenames so I used existing ones.

src/config/app.spec.ts
internalsystemerror: It's normalizing all the versions to these minor version numbers beforehand, so it should in theory:
https://github.com/laminas/laminas-ci-matrix-action/blob/d9110cbc257d7d10cdd4b5345256386c9024c632/src/config/app.ts#L30-L31

Dockerfile
internalsystemerror: typescript all the things \o/

src/config/app.spec.ts
Ocramius: Do we have an existing test for that scenario?

internalsystemerror: Good question, I'm not even sure what's supposed to be returned here, I'm taking the code at face value. Could do with waiting for @boesing (time permitting obviously) to have a look over this, and I can add more tests after that.

Ocramius: I associated many PRs that are affected by this specific problem: they can be seen on the PR overview.

The problem is that a dependency range like `"php": "~8.1.12"` will produce a matrix containing PHP 7.4, 8.0, 8.1, for some reason.

internalsystemerror: Ohhhhhhhhhhhhhhhhh, I see the problem now ... It's that line above, if `composerPhpVersion` is `~8.1.12` then `${version}.0` won't match it. I think that's the issue here.

internalsystemerror: It was the line mentioned above after all. `composerPhpVersion` is the constraint pulled straight from `require.php` in `composer.json`.

internalsystemerror:
Closing in favour of https://github.com/laminas/laminas-ci-matrix-action/pull/139

PR #139, opened 2022-11-01
Fix incorrect version extraction when extracting `php` version from `composer.json` constraints

internalsystemerror:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | yes

### Description

A bug was discovered with several repositories and #136 managed to highlight the issue. This PR proposes a fix for this.
- This PR extends #136.
- ~~One "drawback" which is fine by me, is that since we're comparing `${installableVersion}.99` then pinned constraints such as `8.0.0` won't actually work with it, but at that point, it's unlikely to run using the correct patch anyway, so maybe that shouldn't be supported.~~ See https://github.com/laminas/laminas-ci-matrix-action/pull/139#issuecomment-1300462260 for how this changed.
- This allows `~8.1.12` to use version `8.1`.


internalsystemerror:
Actually this appears to break `<=8.1.0` returning `8.1` :/. Should that be the case? A version of `8.1.12` would not actually match that constraint anyway?

Or maybe it should be if it satisfies `${version.0}` OR if it satisfies `${version}.99`?

weierophinney:
> Actually this appears to break `<=8.1.0` returning `8.1`

I'd prefer that break.

I see no good reason to indicate compatibility with 8.1.0, but not later versions in the 8.1 series.

Ocramius:
Indeed, broken, but "less broken than now" :D

internalsystemerror:
I'll modify the test that fails to indicate that <=8.1.0 should not match the minor 8.1 then.

internalsystemerror:
Actually ... I think I came up with a "better" way, at least in terms of not breaking anything that is currently expecting this operation. I believe the original intention was for the patch number to essentially be ignored, so that's exactly what I've done here.

Any `X.x.x` version is normalised to `X.x.0` and the comparison is then made to `${installableVersion}.0`. This seems to satisfy most of the current functionality whilst also fixing the issues where the patch number > 0.

See https://github.com/laminas/laminas-ci-matrix-action/pull/139/files#diff-7c294c22c7ebe5a7822160fad8ac3789ab3cd7629c7b7637920ef3c72e2cc7b9R7-R20 for how this change is intended to function.

Ocramius (approved):
LGTM :ship:

PR #141, opened 2022-11-02
Bump node to v19 and enforce strict engine checks

internalsystemerror:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| Bugfix        | yes
| BC Break      | no
| New Feature   | yes
| RFC           | no
| QA            | yes

### Description

This PR:
- Closes https://github.com/laminas/laminas-ci-matrix-action/issues/140.
- Enforces strict engine checks causing install to error instead of warn should the wrong node or npm version be used.
- Updates the lock file to include the engine change.

PR #142, opened 2022-11-02
Use correct config key for node-versions

internalsystemerror:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description

This fixes my error from https://github.com/laminas/laminas-ci-matrix-action/pull/141. The correct key is `node-versions` and not just `node` https://github.com/actions/setup-node#supported-version-syntax.


Ocramius:
Weird that it went green in #141 :shrug:

PR #150, opened 2022-11-14
Remove deprecated `set-output` action commands

### LINKED ISSUE #147: GHA `set-output` is deprecated
It seems that the way how we (I) implemented the tests in this project might become invalid at some point in time.
I've recently seen the output of one of the tests within a PR and it says:

> Warning: The `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

I just create this ticket as a reminder that there is a TODO but unfortunately I don't have time as of right now to work on this right now.

gsteel:
`actions/core` is already at `^1.10` so only the workflows require updating to use the new environment files

Closes #147

### DIFF
--- a/.github/workflows/build-and-push-containers.yml
+++ b/.github/workflows/build-and-push-containers.yml
@@ -16,7 +16,7 @@ jobs:
           MAJOR="${PREFIX}:$(echo ${TAG} | cut -d. -f1)"
           MINOR="${MAJOR}.$(echo ${TAG} | cut -d. -f2)"
           PATCH="${PREFIX}:${TAG}"
-          echo "::set-output name=tags::[\"${MAJOR}\",\"${MINOR}\",\"${PATCH}\"]"
+          echo "tags=[\"${MAJOR}\",\"${MINOR}\",\"${PATCH}\"]" >> $GITHUB_OUTPUT
 
       - name: Checkout
         uses: actions/checkout@v3
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -25,7 +25,7 @@ jobs:
         # This provides a dedicated "check" to asynchronously test all integration tests within the tests/ directory
         # The project name from the tests/ directory is then re-used within the "qa" job to run the generated "container"
         # within that directory.
-        run: cd tests; echo "::set-output name=matrix::[\"$(ls -d * | tr '\n' ' ' | sed 's/ $//' | sed 's/ /\",\"/g')\"]"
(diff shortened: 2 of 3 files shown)


Ocramius (approved):
LGTM, but I don't really know what's going on under the hood :D 

Asked for more reviews.

gsteel:
Looks to me that all the remaining notices for `set-output` come from `sergeysova/jq-action@v2` - a patch exists for that here: https://github.com/sergeysova/jq-action/pull/9

samsonasik:
If there is possibly multiline with some escaping, the change seems needs to be something like:

```bash
        run: |
          echo "log<<MESSAGE" >> $GITHUB_OUTPUT
          echo "$(some_data)" >> $GITHUB_OUTPUT
          echo 'MESSAGE' >> $GITHUB_OUTPUT
```

ref 

- https://github.com/phpstan/phpstan-src/pull/1917/files#diff-39c7f1bcfed19e41d56b6260c66af6d03da1257d053b371b7c0de3a3ed01bb13
- https://github.com/rectorphp/rector-src/pull/3016/files

gsteel:
@samsonasik shell-check seems OK with the changed lines - is there something specific that looks wrong? Won't single quotes break the expansions?

samsonasik:
@gsteel probably if there is some special chars like in PHPStan, which when combining with multiline, may break the workflow:

```bash
Run echo "log=$(git log c2ef009c67198d39a6e685a1bca21ceab798d35b..fec8acc849277c78755d3d00767eff95ed3c4014 --reverse --pretty='%H %s' | sed -e 's/^/https:\/\/github.com\/phpstan\/phpstan-src\/commit\//')" >> $GITHUB_OUTPUT
Error: Unable to process file command 'output' successfully.
```

ref https://github.com/phpstan/phpstan-src/actions/runs/3320826153/jobs/5487697232#step:6:1

The multiline example is at https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#example-2

gsteel:
Thanks @samsonasik - I understand now! None of the changed lines output multiline. The CI workflow which dumps a load of json is single line too: https://github.com/laminas/laminas-ci-matrix-action/actions/runs/3460149728/jobs/5776368483#step:3:2

samsonasik (approved):
Ok then 👍

boesing:
Can only tell that tests seem to run, therefore I guess it works for the `continuous-integration` plan.
Not sure if it works properly for the tag stuff regarding container creation, no idea how to test that.

boesing (approved):

internalsystemerror (approved):
Not too familiar with the quirks of GHA, but reading through the changelog, everything here is as they suggest so 👍

Ocramius:
Nobody presses the bad green button? I'll do it then :P

PR #164, opened 2022-12-19
Upgrade npm to v9

internalsystemerror:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | yes
| RFC           | no
| QA            | no

### Description

This PR:
- Updates npm to v9
- Sets the specific version of node to use aswell as npm
- Updates the lockfile with these new versions
- Supersedes https://github.com/laminas/laminas-ci-matrix-action/pull/146
- Closes https://github.com/laminas/laminas-ci-matrix-action/pull/162

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:19-alpine as compiler
+FROM node:19.3.0-alpine as compiler
 
 RUN mkdir -p /usr/local/source
 WORKDIR /usr/local/source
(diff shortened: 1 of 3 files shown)


internalsystemerror:
Odd that all checks ran just fine here?

weierophinney (approved):
Looking over the npm 9 changelog, I don't see anything that would affect us here when upgrading from v8. The main changes were the node ranges it supports (19 is in the upper end of that range), and how it handles and reports error conditions (all of which will only help us).

:ship:

PR #166, opened 2022-12-19
`set-output` deprecation changed how multilines are handled

internalsystemerror:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description

Several tags were being created that were invalid e.g. `v1%0A1.21.0`. It seems that when set-output was deprecated, github changed how multiline strings were handled. This PR should fix that issue according to: https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#multiline-strings

### DIFF
--- a/.github/workflows/create-additional-action-tags.yml
+++ b/.github/workflows/create-additional-action-tags.yml
@@ -16,7 +16,7 @@ jobs:
           TAG=${GITHUB_REF/refs\/tags\//}
           MAJOR="v$(echo ${TAG} | cut -d. -f1)"
           MINOR="${MAJOR}.$(echo ${GITHUB_REF} | cut -d. -f2)"
-          EOF="$(uuid)"
+          EOF="$(uuidgen)"
           echo "tags<<${EOF}" >> $GITHUB_OUTPUT
           echo "${MAJOR}" >> $GITHUB_OUTPUT
           echo "${MINOR}" >> $GITHUB_OUTPUT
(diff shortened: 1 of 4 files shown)


internalsystemerror:
Rebased using https://github.com/laminas/laminas-ci-matrix-action/pull/164 to allow CI to pass

.github/workflows/create-additional-action-tags.yml
```diff
@@ -16,7 +16,9 @@ jobs:
           TAG=${GITHUB_REF/refs\/tags\//}
           MAJOR="v$(echo ${TAG} | cut -d. -f1)"
           MINOR="${MAJOR}.$(echo ${GITHUB_REF} | cut -d. -f2)"
-          echo "tags=${MAJOR}%0A${MINOR}" >> $GITHUB_OUTPUT
+          echo "tags=${MAJOR}<<EOF" >> $GITHUB_OUTPUT
+          echo "${MINOR}" >> $GITHUB_OUTPUT
+          echo "EOF" >> $GITHUB_OUTPUT
```
Xerkus: I think this should look like this:
```suggestion
          echo "tags<<EOF" >> $GITHUB_OUTPUT
          echo "${MAJOR}" >> $GITHUB_OUTPUT
          echo "${MINOR}" >> $GITHUB_OUTPUT
          echo "EOF" >> $GITHUB_OUTPUT
```

Xerkus (approved):

weierophinney (approved):
Discussed in slack; changes look good. :ship:

Ocramius:
Nasty one: good catch!

PR #167, opened 2022-12-19
Diff variable needs to be unquoted when passed to the application

internalsystemerror:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description

Sometimes when running the action on a pull request, it's unable to correctly perform the regex checks on the list of filenames and incorrectly returns no checks to be performed (see https://github.com/laminas/laminas-filter/actions/runs/3732216426/jobs/6332977468).

This appears to have been introduced by https://github.com/laminas/laminas-ci-matrix-action/pull/117 and after various discussion over slack and with @Xerkus help, it is believed that this PR should fix that issue.

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -80,4 +80,6 @@ if [[ "$DIFF" != "" ]];then
     echo "${DIFF}"
 fi
 
-/action/main.js "${DIFF}"
+# This variable is deliberately unquoted so that it can be properly processed on the other side.
+# shellcheck disable=SC2086
+/action/main.js ${DIFF}


Xerkus (approved):

entrypoint.sh
```diff
@@ -80,4 +80,6 @@ if [[ "$DIFF" != "" ]];then
     echo "${DIFF}"
 fi
 
-/action/main.js "${DIFF}"
+# This variable is deliberately unquoted so that it can be properly processed on the other side.
+# shellcheck disable=SC2086
+/action/main.js ${DIFF}
```
weierophinney, line 85: That is sooooo odd; normal rule of thumb is to ALWAYS quote arguments in shell scripts. Would love to know why this occurs...

internalsystemerror: Same. I think the list of files gets read as a single string, but I'm not certain. It should be easy enough to see what's happening around https://github.com/laminas/laminas-ci-matrix-action/blob/5e50256611e353e2f4f002caf1d99a631889e97f/src/main.ts#L14-L22 though.

Xerkus: Each file name from diff is expected to be a separate argument and it explicitly relies on shell word splitting behavior using IFS.
Double quoting prevented word splitting and the whole diff ended up passed as a single argument which then failed to match non-multiline regex like `\.php$` as it would effectively be matching only the very last file

PR #184, opened 2023-02-08
Extend `exclude` with `php` and `dependencies` settings

### LINKED ISSUE #93: Allow `php` and `dependency` in `exclude` along with `name` reflecting the tools/checks name
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| BC Break    | no

#### Summary

Currently, one has to create plenty of `exclude` for PHPUnit for example.
Given the fact, that PHPUnit is executed in at least `lowest` and `latest` (and also `locked` for minimum supported PHP version), that would mean at least 3 entries within `exclude` to exclude a specific PHP version with all dependencies.

I'd like to suggest something like this (on top of the current way of excluding checks):

```json
{
     "exclude": [
        {
             "name": "PHPUnit",
             "php": "8.0",
             "dependencies": "*",
        }
    ]
}
```

This would exclude all PHPUnit runs on PHP 8.0 with all composer dependencies.

boesing:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| New Feature   | yes

### Description

This introduces the ability to exclude specific jobs for specific PHP and/or composer dependency sets.
Fixes #93


README.md
```diff
@@ -200,18 +200,45 @@ The tool discovers checks first, then appends any `additional_checks` are concat
...
+   "exclude": [
+      {
+         "name": "PHPUnit",
+         "php": "8.0"
+      }
+   ]
+}
+```
+
+In case you only want to exclude jobs for specific composer dependency sets, add `dependencies` to the `exclude` configuration:
```
froschdesign: "Composer" is a proper name and for these names we respect the correct notation.

```suggestion
In case you only want to exclude jobs for specific Composer dependency sets, add `dependencies` to the `exclude` configuration:
```

README.md
```diff
@@ -200,18 +200,45 @@ The tool discovers checks first, then appends any `additional_checks` are concat
 
 ### Excluding specific jobs
 
-The easiest way to exclude a single job is via the `name` parameter:
+You can exclude specific jobs by using their names:
```
froschdesign, line 203: For killing the word "easiest" 👍🏻 

Otherwise nice additions with examples to the different options.

Ocramius:
@internalsystemerror is the CI failure related to what you fixed in #190 ?

internalsystemerror:
I believe you mean @gsteel although I see it losing the "before_script" from the matrix output for some reason.

I did some similar changes in https://github.com/laminas/laminas-ci-matrix-action/pull/167 and https://github.com/laminas/laminas-ci-matrix-action/pull/166 but upon reviewing those, I don't think they would affect this either.

internalsystemerror:
It looks like the only issue here is that it's no longer adding the `xmllint` checks in the matrix output whilst that is expected. Given the only recent change is #190, I suspect the top change (the one I commented on) needs reverting and adding `multiline: true` is all that's required.

README.md
```diff
@@ -200,18 +200,45 @@ The tool discovers checks first, then appends any `additional_checks` are concat
...
 
 ```json
 {
   "exclude": [
     {
-      "name": "PHPUnit on PHP 8.0 with latest dependencies"
+      "name": "PHPUnit"
```
Ocramius, line 209: Will the old exclusion via exact name match still work? I saw no changes in tests, which seems to hint at that, but I'm not 100% sure that's covered :D

Ocramius (approved):
LGTM :+1: 

@boesing merge at your discretion after checking my last question :+1: 

We can also tag a new minor, before going to a new major with #189

README.md
boesing: yes that would work but something is still not properly working and thus have to re-check here. No need to have this as a blocker for v2. I can adapt.

Ocramius: Nono, I mean that you can merge and release `1.23.x` whenever: I'll deal with `2.0.x` over the weekend, if I can.

Ocramius:
@boesing @internalsystemerror should I rebase and release this to `1.23.x`, if green?

boesing:
It is not properly working yet, please ignore this for any release. I'll handle this once I find some time to work on this. There is no pressure on this topic and no high need (at least for me) as of now 👍🏼

Ocramius:
Ok, just beware that it will go into `2.1.x`, probably.

PR #190, opened 2023-02-22
Prevent multiline matrix payload from causing issues in GitHub workflow output

gsteel:
Makes CI green.

It appears that sergeysova/jq-action@v2.3.0 was at fault here, and enabling its `multiline` option solves the issue.

This patch also updates usage of `$GITHUB_OUTPUT` in `continuous-integration.yml` in case the matrix generated there is also multiline.

Deprecations are still being issued about `set-output` but I'm not sure where these are coming from 🤷‍♂️

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -91,12 +91,12 @@ jobs:
         run: cd tests/${PROJECT_NAME_TO_TEST} && docker run -i --entrypoint "/action/main.js" -v $(realpath .):/github/workspace -w=/github/workspace ${TEST_TAG} $(test -r diff && cat diff || echo -n "")
 
       - name: "Output generated matrix"
-        uses: sergeysova/jq-action@v2
+        uses: sergeysova/jq-action@v2.2.1
         with:
             cmd: "jq . < <(echo '${{ steps.matrix_generation.outputs.matrix }}')"
 
       - name: Minify matrix from test directory
-        uses: sergeysova/jq-action@v2
+        uses: sergeysova/jq-action@v2.2.1
         id: expected_matrix
         env:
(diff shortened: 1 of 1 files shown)


.github/workflows/continuous-integration.yml
```diff
@@ -25,7 +25,11 @@ jobs:
         # This provides a dedicated "check" to asynchronously test all integration tests within the tests/ directory
         # The project name from the tests/ directory is then re-used within the "qa" job to run the generated "container"
         # within that directory.
-        run: cd tests; echo "matrix=[\"$(ls -d * | tr '\n' ' ' | sed 's/ $//' | sed 's/ /\",\"/g')\"]" >> $GITHUB_OUTPUT
+        run: |
```
internalsystemerror, line 28: Is this needed ASWELL as pinning the version of `jq-action`? I would hope this change would be all that's needed?

gsteel: I don't think so - But, if there's any chance that `matrix` _is_ multiline, at least the case is covered. It should probably always have been like this WRT #150 

We may be able to stick with `jq-action@v2` if the `multiline` option works out for us.

Ocramius (approved):
Thanks @gsteel!

.github/workflows/continuous-integration.yml
internalsystemerror: I think this change needs reverting as `"matrix=[\"$(ls -d * | tr '\n' ' ' | sed 's/ $//' | sed 's/ /\",\"/g')\"]"` should be on a single line.

gsteel: I'll send in another patch @internalsystemerror for 1.23

internalsystemerror: Ty <3

PR #191, opened 2023-02-22
Revert multiline changes to `continuous-integration.yml`

gsteel:
The issue fixed in #190 is fixed due to adding the `multiline` option to `sergeysova/jq-action` - The changes to `continuous-integration.yml` are spurious because the JSON payload is already a single line.


internalsystemerror (approved):
Specifically, I think `tr '\n' ' '` is already handling it.

internalsystemerror:
Thanks @gsteel

PR #196, opened 2023-03-02
Cleanup/remove EOL PHP versions from matrix

Ocramius:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | yes
| New Feature   | no
| RFC           | no
| QA            | yes

Ref: https://github.com/laminas/laminas-continuous-integration-action/issues/149


internalsystemerror (approved):
Dropping support in the next major version. No objections, I too like to **\o/ burn all the things \o/** but what's the motivation? I know v1 will always be there, but is having these older versions around that much of a burden?

Ocramius:
Main motivation is that packages upgrading to v2 now should also stop supporting PHP 7.

Secondary: reduce container size and clutter 👍

lcobucci (approved):
:ok_hand:

PR #226, opened 2023-07-09
Verify BC breakages on pull requests, by comparing PR against base branch

boesing:
|    Q          |   A
|-------------- | ------
| New Feature   | yes

### Description

This PR continues the work of #126 while adapting some changes to reduce complexity:

- reduces event handling and focusses on `GITHUB_BASE_REF` environment variable instead
- introduces a new Tool type `ToolRunningContainerDefaultPhpVersion` which targets tools which are pre-installed within the CI container (such as `roave-backward-compatibility-check`), it provides `@default` via `php` job configuration and thus, the CI container won't change the PHP version which is the default by the moment of executing the container (needs https://github.com/laminas/laminas-continuous-integration-action/pull/158)
- Provides a configuration flag `backwardCompatibilityCheck` to opt-in for the `roave-backward-compatibility-check`, with v2 of the CI pipeline, this flag might be changed to be enabled by-default and projects need to opt-out for it

@Xerkus @internalsystemerror I've added you guys as you gave feedback in the initial #126 PR.
@Ocramius I hope its okay that I create a dedicated PR and supersede #126 as I am unable to push to your fork.


tests/code-check-roave-backward-compatibility/matrix.json
```diff
@@ -0,0 +1,10 @@
+{
+    "include": [
+        {
+            "name": "Backward Compatibility Check [@default, latest]",
+            "job": "{\"command\":\"roave-backward-compatibility-check check --from=\\\"1111222233334444aaaabbbbccccdddd\\\" --install-development-dependencies\",\"php\":\"@default\",\"extensions\":[],\"ini\":[],\"dependencies\":\"latest\",\"ignore_platform_reqs_8\":false,\"ignore_php_platform_requirement\":false,\"additional_composer_arguments\":[],\"before_script\":[]}",
```
Ocramius, line 5: `--install-development-dependencies` should probably be dropped, but OK this way for a first iteration

laminas-ci.schema.json
```diff
@@ -328,14 +331,19 @@
...
+            "description": "This PHP version is used for all QA check jobs. The default depends on the `composer.json` of the project and usually reflects the minimum supported PHP version of that project.",
+            "examples": [
+                "8.0"
+            ]
+        },
+        "backwardCompatibilityCheck": {
+            "type": "boolean",
+            "title": "Flag to enable/disable backwards compatibility check",
+            "description": "This flag enables/disables backwards compatibility check using roave/backward-compatibility-check."
+        },
```
Ocramius: Should we declare a default, so it is self-documenting on that?

src/config/app.spec.ts
```diff
@@ -47,7 +56,41 @@ describe('config/app', () => {
...
                 additionalComposerArguments   : [],
+                backwardCompatibilityCheck    : false,
+                targetReference               : null,
+            });
+        });
+
+        it('should detect GITHUB_BASE_REF', () => {
+            const environment = process.env;
+
+            configDotenv({path: `${roaveBackwardCompatibilityPath}/test.env`});
```
Ocramius, line 67: Fancy solution :O

src/config/app.spec.ts
```diff
@@ -47,7 +56,41 @@ describe('config/app', () => {
...
+                stablePhpVersion              : '7.4',
+                minimumPhpVersion             : '7.4',
+                latestPhpVersion              : '7.4',
+                lockedDependenciesExists      : false,
+                phpExtensions                 : [],
+                phpIni                        : [],
+                ignorePhpPlatformRequirements : {},
+                additionalComposerArguments   : [],
+                backwardCompatibilityCheck    : true,
+                targetReference               : '1111222233334444aaaabbbbccccdddd',
```
Ocramius: More than the "target" reference, this is a "base" reference.

Perhaps a more clear name would improve things a bit

tests/code-check-roave-backward-compatibility/matrix.json
Ocramius, line 5: This uses `\"php\":\"@default\"`, but `roave/backward-compatibility-check` generally runs on very new versions.

I suppose this PHP version requiorement can be baked in the `roave-backward-compatibility-check` shell command: will check the companion PR to this.

Ocramius (approved):

laminas-ci.schema.json
boesing: Is that only `"default": false`?
If so, ofc we can 👍🏼

src/config/app.spec.ts
boesing: seems to be kinda standard for JEST, had only a quick search on google and it was the first suggestion once it comes to testing stuff including environment 😅

laminas-ci.schema.json
Ocramius: Yep

src/config/app.spec.ts
boesing: I can rename it to `baseReference` if thats more clear?

tests/code-check-roave-backward-compatibility/matrix.json
boesing: > This uses \"php\":\"@default\", but roave/backward-compatibility-check generally runs on very new versions.

Even tho there might be a version which might be compatible, the shell command will be installed while container is being created. Due to this, a specific version of `roave/backward-compatibility-check` is checked out. Pretty sure that if I check the latest version out, it does **not** work with PHP 5.6.
If it does, maybe you should allow PHP 5.6 within `composer.json`.

boesing: I just kept it as is, I can remove it if it makes no sense. 🤷🏼‍♂️ 
I have no clue what it does 😅

src/config/app.spec.ts
Ocramius: Works for me!

tests/code-check-roave-backward-compatibility/matrix.json
Ocramius: I did not fully understand your last comment, so perhaps I'm misunderstanding what this does? :thinking:

Ocramius: The default behavior is `composer install --no-dev` while running the tool, so it will assume that any sources in `"autoload"` can be analyzed.

With this flag, it will remove the `--no-dev`, and run a full installation, with dev dependencies.

Thinking of it further, I believe we don't currently care that much: let's leave it there, for now :D

boesing: when the CI container is created and released, a default php Version is set. with the PR regarding roave bc check, I added composer.json to the root with a php Version compatible with all the Tools we do require, including roave bc check. This php Version is then set as the default PLUS runs composer install. 
It will remain as the default php Version. I do not have any clue in this container here, what default Version is executed in the container and thus, I am fine with having `@default` to ensure that whatever default version is set today and in 10 years, that it can run the bc check tool.
Since I do have to pass a php Version as it is required by the container and such, I rather give it a meaningful Name rather than sending 5.6 but the Tool is executed with 8.1 afterwards. 

the composer.lock in the CI container installs bc Tool which only allows php 8.1 and 8.2 and thus it is not compatible with any PHP Version or I totally miss understand what the php Version requirements are used for in that package.

So to not have to worry about Whatever php Version, I just Trust the container to run a php Version which can execute whatever bc Tool Version was installed during container build.

Ocramius: Related to https://github.com/laminas/laminas-continuous-integration-action/pull/158#discussion_r1259450707

I'm fine with the current approach: let's try and roll with it :)

PR #232, opened 2023-07-11
Use proper escape sequence for passing the base reference to `roave-backward-compatibility-check`

boesing:
|    Q          |   A
|-------------- | ------
| Bugfix        | yes

### Description

Hopefully this fixes the problem with the CI container which receives unescaped `--from` option value. 🤞🏼

### DIFF
--- a/src/tools.ts
+++ b/src/tools.ts
@@ -56,10 +56,9 @@ function backwardCompatibilityCheckTool(config: Config): ToolRunningContainerDef
     }
 
     return {
-        // @TODO need to `git fetch baseSha1` from source repo!
         executionType : ToolExecutionType.STATIC,
         name          : 'Backward Compatibility Check',
-        command       : `roave-backward-compatibility-check check --from="${ config.baseReference }" --install-development-dependencies`,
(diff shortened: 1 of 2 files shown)


src/tools.ts
```diff
@@ -56,10 +56,9 @@ function backwardCompatibilityCheckTool(config: Config): ToolRunningContainerDef
     }
 
     return {
-        // @TODO need to `git fetch baseSha1` from source repo!
         executionType : ToolExecutionType.STATIC,
         name          : 'Backward Compatibility Check',
-        command       : `roave-backward-compatibility-check check --from="${ config.baseReference }" --install-development-dependencies`,
+        command       : `roave-backward-compatibility-check check --from=\\"${ config.baseReference }\\" --install-development-dependencies`,
         filesToCheck  : [ 'composer.json' ],
```
ghostwriter, line 62: we should use a single backslash;

```diff
- --from=\\"${ config.baseReference }\\"
+ --from=\"${ config.baseReference }\"
``` 

or drop the quotes:

```diff
- --from=\\"${ config.baseReference }\\"
+ --from=${ config.baseReference }
```

PR #292, opened 2024-04-23
Enable backward compatibility check for all laminas related organization checks

boesing:
|    Q          |   A
|-------------- | ------
| New Feature   | yes

### Description

Unless explicitly disabled in a project, backward compatibility checks are always executed for all laminas related organization checks.

### DIFF
--- a/src/config/app.ts
+++ b/src/config/app.ts
@@ -473,6 +473,14 @@ export default function createConfig(
         baseReference = null;
     }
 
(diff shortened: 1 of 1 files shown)


boesing:
Well, that would've been too easy.
I guess `GITHUB_REPOSITORY_OWNER` does not reflect the targeted repository but the source repository.

Not sure if I will continue work on this in the near future. In case s1 might find a solution, just let me know.

PR #298, opened 2024-04-24
Synchronize `package.json` and `package-lock.json` with `Dockerfile` version of `node`

boesing:
|    Q          |   A
|-------------- | ------
| QA            | yes

### Description

The `Dockerfile` got updated in #285 but the `package.json` and `package-lock.json` were not updated.


boesing:
Hm, still have to verify that both `node` version of `Dockerfile` is in sync. I've checked some stuff and it felt better to have a generic package-lock.json check but this will still not prevent `Dockerfile` from diverging from `package.json` if that is never updated...

WIll have another look tomorrow, its too late now...

PR #314, opened 2024-09-10
Fix Incorrect globbing of docs directory during linting checks

gsteel:
|    Q          |   A
|-------------- | ------
| Bugfix        | yes

### Description

Markdown linting is not globbing all the files expected in the doc/book tree - perhaps the binary should be markdownlint-cli2 which is what's installed - and possibly has deps on different globbing libs that work???

Example run where 12 files are matched instead of expected 123:
https://github.com/laminas/laminas-validator/actions/runs/10796875119/job/29946873936?pr=390#step:3:286

### DIFF
--- a/src/tools.ts
+++ b/src/tools.ts
@@ -70,14 +70,14 @@ export default function createTools(config: Config): Array<Tool> {
         {
             executionType : ToolExecutionType.STATIC,
             name          : 'Documentation Linting',
-            command       : 'markdownlint doc/book/**/*.md',
+            command       : "markdownlint 'doc/book/**/*.md'",
             filesToCheck  : [ 'doc/book/' ],
             toolType      : ToolType.LINTER,
         },
         {
             executionType : ToolExecutionType.STATIC,
             name          : 'Documentation Linting',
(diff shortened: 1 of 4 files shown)


gsteel:
So, the problem here is not the tool path - `laminas-continuous-integration-action` aliases `markdownlint-cli2` to `markdownlint` during build.

The problem is the expansion of the command argument in the shell.

Currently, the job receives

```shell
markdownlint docs/book/**/*.md
```

where what it actually needs is

```shell
markdownlint 'docs/book/**/*.md'
```

Xerkus:
So the tool is doing globbing internally?

I do not see why would it not work with shell expansions. Number of arguments it can handle I guess?

Good catch.

gsteel:
I should probably rebase onto 1.27.x

Xerkus:
Oh boy. Here come all the documentation lint failures again.

Xerkus (approved):

gsteel:
Post-Release, all 123 files in laminas-validator are detected now:
https://github.com/laminas/laminas-validator/actions/runs/10801018942/job/29960270516?pr=392

PR #341, opened 2025-10-16
Upgrade Node Dependencies

gsteel:
Required removal of [eslint-config-incredible](https://github.com/pustovitDmytro/eslint-config-incredible) during upgrade to eslint v9

Adds a simple Makefile to make it easier to muck about with JS deps via docker

- Closes #337
- Closes #336
- Closes #335
- Closes #334
- Closes #329
- Closes #309
- Closes #308
- Closes #300
- Closes #296


Makefile
```diff
@@ -0,0 +1,22 @@
+
+NODE_IMG=node:24.10.0-alpine
```
Ocramius: Perhaps we should reference a base layer in our Dockerfile, so this is always in sync?

Ocramius (approved):

Makefile
gsteel: How would I do that?

Ocramius: Would meet to build+tag an image with `--target=`, then use that tag to run the remaining steps

gsteel: I think I understand - I've adjusted `Makefile` to build and tag the image locally, but `docker build --target=compiler` just builds everything anyway - not too familiar with multi-stage builds. At least what I've done will stop node versions drifting apart

Ocramius (approved):
Excellent, thanks @gsteel

PR #358, opened 2025-11-23
Add support `shipmonk/composer-dependency-analyser`

sasezaki:
|    Q          |   A
|-------------- | ------
| New Feature   | yes

### Description

I propose to add `shipmonk/composer-dependency-analyser
https://github.com/shipmonk-rnd/composer-dependency-analyser

* It has already been added to CI in the v5 branch of laminas-hydrator
  * https://github.com/laminas/laminas-hydrator/pull/136
  * However, currently the command is executed as `composer-dependency-analyser --disable-ext-analysis`, so the configuration file needs to be modified to $config->disableExtensionsAnalysis()

* I'm also using it personally (in work projects), and I think it's functioning excellently


gsteel (approved):
This looks good to me @sasezaki Thank you :)

I'll wait for others to check it over too however as I'm not 100% on this codebase 😬

Ocramius (changes requested):
As a maintainer of `maglnet/composer-require-checker`, I'm vary of this.

The reason is that I know that the shipmonk dependency cuts corners aggressively there, skipping the AST world, and diving directly into the "let's parse code with regexes". EDIT: [historic reference](https://stackoverflow.com/a/1732454/347063)

I'd rather see improvements in tooling that does things formally, than introducing this specific tool 🤔 

EDIT: I stand corrected - it uses reflection. I'm not 100% sure about this then, will need to check again.

sasezaki:
Hi @Ocramius 

I believe your concern is whether we should adopt this as the de facto standard for composer dependency checking in the Laminas project.

I don't intend this PR to replace composer-require-checker. 
I think it would be good if this could be used as a secondary checker for the time being, while keeping composer-require-checker as the primary tool. As mentioned in the hydrator issue, detecting "unused" dependencies can also serve as a clue for reviewing dependencies.

Ocramius (approved):
Sounds good to me: I did a new check over this package, and I believe it changed substantially since last time I investigated it.

Can agree with @sasezaki's point 👍

PR #371, opened 2026-02-04
Add working-directory input for monorepo support

guillaume-sainthillier:
## Summary
This PR adds support for monorepo configurations by introducing a `working-directory` input parameter that allows the action to operate within a specific subdirectory of the repository.

## Key Changes
- **New Input Parameter**: Added `working-directory` input to `action.yml` with a default value of `.` (current directory)
  - Allows users to specify a subdirectory where `composer.json` and other configuration files are located
  - Marked as optional to maintain backward compatibility

- **Diff Filtering Logic**: Implemented filtering in `entrypoint.sh` to:
  - Extract only files that exist within the specified working directory
  - Strip the working directory prefix from file paths to make them relative
  - Provide clear logging of filtered results

- **Directory Navigation**: Added logic to change into the working directory before executing the main action script
  - Ensures all subsequent operations (composer analysis, etc.) run in the correct context
  - Includes error handling if directory change fails

## Implementation Details
- The `WORKING_DIR` variable defaults to `.` if `INPUT_WORKING_DIRECTORY` is not set, preserving existing behavior
- Diff filtering only applies when a non-default working directory is specified
- File paths are compared with the working directory prefix to ensure accurate filtering
- Clear console output indicates when filtering occurs and reports which files are affected

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -155,6 +155,13 @@ jobs:
       - name: Run Jest
         run: "npm run test"
 
+  shell-testing:
+    runs-on: "ubuntu-latest"
+    steps:
+      - uses: "actions/checkout@v6"
+      - name: Run shell tests
+        run: bash scripts/entrypoint.spec.sh
+
   package-lock-up-to-date:
     runs-on: "ubuntu-latest"
     steps:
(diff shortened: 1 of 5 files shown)


guillaume-sainthillier:
hello @gsteel @Ocramius, any news about this PR ? something to change for merging ?
We have several repositories using react / php (`client` / `server` directories), it would be great to make it works natively :)

PR #384, opened 2026-03-15
draft : Enable `backwardCompatibilityCheck` by default for Laminas-related repositories

fezfez:
attempt to fix https://github.com/laminas/laminas-ci-matrix-action/pull/292


src/config/app.ts
```diff
@@ -29,6 +29,11 @@ import {
 
 export const OPERATING_SYSTEM = 'ubuntu-latest';
 export const ACTION = 'laminas/laminas-continuous-integration-action@v1';
+export const LAMINAS_RELATED_ORGANIZATIONS = [
+    'laminas',
+    'mezzio',
+    'laminas-api-tools',
+];
```
froschdesign, line 36: I think this information shouldn’t be included here, as the action is not only used in our packages but is intended to be used generally and independently.
Regardless of that, the API tools are no longer relevant.

PR #387, opened 2026-03-29
fix: redirect stderr before cs2pr to prevent PHP deprecation warnings from breaking XML parsing

guillaume-sainthillier:
## Problem

On PHP 8.5, packages like `sentry-symfony` emit deprecation warnings (`__sleep()/__wakeup()` serialization magic methods deprecated). When tools output checkstyle XML and pipe through `cs2pr`, these warnings get mixed into stdout, causing `cs2pr` to fail with:

```
Error: Start tag expected, '<' not found on line 2, column 1
```

This affects all tools using `| cs2pr`: PHPCodeSniffer, PHP CS Fixer, and Twig CS Fixer.

## Solution

Add `2>/dev/null` before the `| cs2pr` pipe to redirect stderr, ensuring only clean XML reaches `cs2pr`.

This is a minimal, non-breaking change that also future-proofs against any other deprecation warnings from third-party packages.

## Changes

- `phpcs`: `./vendor/bin/phpcs -q --report=checkstyle 2>/dev/null | cs2pr`
- `php-cs-fixer`: `./vendor/bin/php-cs-fixer fix -v --diff --dry-run --format=checkstyle 2>/dev/null | cs2pr`
- `twig-cs-fixer`: `./vendor/bin/twig-cs-fixer lint --report checkstyle 2>/dev/null | cs2pr`
- Updated all corresponding test fixtures

### DIFF
--- a/src/tools/codesniffer.ts
+++ b/src/tools/codesniffer.ts
@@ -4,7 +4,7 @@ import {ToolExecutionType} from '../enum/toolExecutionType';
 export const PhpCodeSnifferTool = {
     executionType     : ToolExecutionType.STATIC,
     name              : 'PHPCodeSniffer',
-    command           : './vendor/bin/phpcs -q --report=checkstyle | cs2pr',
+    command           : './vendor/bin/phpcs -q --report=checkstyle 2>/dev/null | cs2pr',
     filesToCheck      : [ 'phpcs.xml', 'phpcs.xml.dist' ],
     toolType          : ToolType.CODE_CHECK,
     lintConfigCommand : 'xmllint --schema vendor/squizlabs/php_codesniffer/phpcs.xsd',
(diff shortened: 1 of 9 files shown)


gsteel (approved):
Thank you @guillaume-sainthillier - LGTM, but I will wait for another approving review from the team before merging 👍

Ocramius (changes requested):
For me, this change is **not OK**: it redirects STDERR to `/dev/null`, and I believe something else is wrong instead.

In theory, the `|` operator shouldn't be directing STDERR to the next tool anyway, so this needs to be re-checked.

Suppression of any tool's STDERR is not viable in any CI: it's a massive problem for later on, should anything else happen / require investigation.

guillaume-sainthillier:
@Ocramius I agree with your decision, and not able to remember what project was causing this issue :D closing this until remember the use case

Issue #395, opened 2026-08-03
Add Support for Mago


gsteel:
### Feature Request

Add support for running:

- `mago lint` https://mago.carthage.software/1.45.0/en/tools/linter/usage/#lint-the-whole-project
- `mago analyse` https://mago.carthage.software/1.45.0/en/tools/analyzer/command-reference/
- `mago fmt --check` https://mago.carthage.software/1.45.0/en/tools/formatter/usage/#ci-check-without-rewriting

Each of these built-in tools have a `--minimum-fail-level` switch that has a counterpart configuration option, so I think that further configuration is not necessary given that users can specify this in `./mago.toml`

So, I think that it is sufficient to detect:

- The presence of `carthage-software/mago` in `composer.json`
- The presence of `./mago.toml`

and run the basic commands as they stand.

### Possible Issues

Users may not be using all tools, for example, I may wish to use the "analyser" and linter but not the formatter, therefore running all 3 tools when Mago is detected may be problematic if you favour say `PHP_CodeSniffer` for CS rules.

To allow for this case, it might be helpful to add switches to `.laminas-ci.json` such as

```json
{
    "mago-skip-lint": bool,
    "mago-skip-fmt": bool,
    "mago-skip-analyse": bool,
}
```

It is rumoured that the `lint` and `analyse` commands might be collapsed into a single command in v2.0 _(probably just `analyse`)_, so we may need to internally detect the version of Mago in use to account for that either now or in future.

Note that Mago uses `analyse` and `analyze` interchangeably.

CC @weierophinney

gsteel:
Closed via #399

PR #399, opened 2026-08-03
Initial support for Mago

gsteel:
See #395 

I figured that users can use the already available `"exclude": []` option to exclude `lint` if it becomes necessary.


boesing (approved):
Could you also add this tool to the `README.md`?
AFAIK, we list all (hopefully) tools there in some kind of overview.

